PR #147, opened 2014-12-22
Add support for eldoc-mode.

ikirill:
eldoc-mode displays docstrings in minibuffer for symbols and function
calls under point. Implemented by using existing completion functions,
which already return documentation and types.

To use: enable eldoc and irony-eldoc, place point at a symbol or inside
a function call, eldoc will show a message in minibuffer showing type
and, for calls, the function argument name+type in the type signature
will be highlighted.

Notes:

I use irony-mode in C++, and the way I get eldoc right now is with
c-eldoc. But irony-mode already has all the information it needs to
provide eldoc support, since docstrings (?) and especially types are
supplied by the auto-completion functions. irony-mode also, I think,
has the better support for compilation options, and also benefits from
not reparsing files as much.

So since the feature comes almost for free, I think it should be added
even though it might duplicate some other packages.

It works fine, and does not crash on the things that I've tried. Some
things are not implemented, such as displaying types for overloaded
functions where types of arguments matter.

I enabled this by default in irony-mode.

Thanks for irony-mode.

Kirill

### DIFF
--- a/irony-eldoc.el
+++ b/irony-eldoc.el
@@ -1,0 +1,331 @@
+;; -*- lexical-binding: t -*-
+;;; irony-eldoc --- irony-mode support for eldoc
+;;
+;; Copyright (C) 2014 Kirill Ignatiev <github.com/ikirill>
+;;
+;; This program is free software; you can redistribute it and/or modify
+;; it under the terms of the GNU General Public License as published by
+;; the Free Software Foundation, either version 3 of the License, or
+;; (at your option) any later version.
+;;
+;; This program is distributed in the hope that it will be useful,
+;; but WITHOUT ANY WARRANTY; without even the implied warranty of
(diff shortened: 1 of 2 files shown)


Sarcasm:
I will think about avoiding unnecessary reparsing requests. Improving performances is not on the top of my list but this is something I have in mind.

As for irony-eldoc, I will try to put a link to it in irony-mode's documentation/"feature showcase", I will keep the issue open until it's done.

If you want to me to "review" irony-eldoc let me know, for now I will content myself to use it and report things that bothers me from the user-point of view, if any.

Sarcasm:
The README now mentions irony-eldoc: https://github.com/Sarcasm/irony-mode/commit/cf02d4abe2619d921f03011337f04f9951c94ab5

PR #153, opened 2014-12-31
Implement "jumping to declaration" on irony-server.

xuhdev:
Jump to the declaration of the symbol on a specific location.

```
Usage: irony-server find-declaration file-path line col

Output: ("path-to-file-with-declaration" line col)
```

Besides it can jump to the definition of some symbols, such as macros, it's also useful to find out which header file contains declaration of a symbol.

### DIFF
--- a/server/src/Command.cpp
+++ b/server/src/Command.cpp
@@ -152,6 +152,7 @@ Command *CommandParser::parse(const std::vector<std::string> &argv) {
     break;
 
   case Command::Complete:
+  case Command::FindDeclaration:
     positionalArgs.push_back(StringConverter(&command_.file));
     positionalArgs.push_back(UnsignedIntConverter(&command_.line));
     positionalArgs.push_back(UnsignedIntConverter(&command_.column));
--- a/server/src/Commands.def
+++ b/server/src/Commands.def
@@ -16,6 +16,9 @@ X(Diagnostics, "diagnostics", "FILE - look for diagnostics in file")
 X(Complete,
   "complete",
   "FILE LINE COL - perform code completion at a given location")
+X(FindDeclaration,
+  "find-declaration",
+  "FILE LINE COL - find the declaration of the symbol at a given location")
 X(Help, "help", "show this message")
 X(Exit, "exit", "exit interactive mode, print nothing")
 X(SetDebug, "set-debug", "[on|off] - enable or disable verbose logging")
--- a/server/src/Irony.cpp
+++ b/server/src/Irony.cpp
@@ -342,3 +342,40 @@ void Irony::complete(const std::string &file,
     std::cout << ")\n";
   }
 }
+
+void Irony::findDeclaration(const std::string &file,
+                            unsigned line,
+                            unsigned col,
+                            const std::vector<std::string> &flags,
+                            const std::vector<CXUnsavedFile> &unsavedFiles) {
+
+  CXTranslationUnit tu = tuManager_.getOrCreateTU(file, flags, unsavedFiles);
+
+  if (tu == nullptr) {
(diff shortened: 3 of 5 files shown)


Sarcasm:
Here again I'm doubtful about the *perfectness* of this method.

What this returns for forward declaration?

If I have

```cpp
class Foo; // in a header file

// later in a cpp file

void bar(Foo *f);
```

Looking for the declaration of `Foo` in the bar method will return what?

What happens if there are multiple forward declarations? Which one is picked?

What does is return on out-of-line definitions of methods? The docs says:

> If the input cursor is a declaration or definition, it returns that declaration or definition unchanged.

xuhdev:
Yes it could take me quite a while to implement the elisp part, as I'm not too familiar with elisp. But we can use the server command line to test this feature any way.

I agree that this method is not perfect, but is useful in most cases to find out the prototype, or the header file which contains the prototype. If one really needs a perfect feature to jump to declaration, they can use rtags, which has a disadvantage of not working on Windows.

Sarcasm:
> Yes it could take me quite a while to implement the elisp part, as I'm not too familiar with elisp. But we can use the server command line to test this feature any way.

The thing is this is not the most useful feature to me. Sure I can certainly bind all the libclang functions into irony-server but libclang is well tested already, I don't need to test it more than that. What is needed is some well integrated logic into Emacs. I have a few already on my mind, so I'm not too keen to add an "un-perfect" feature, without the proper Elisp binding I'm sorry.

You said:

> I agree that this method is not perfect, but is useful in most cases to **find out** the prototype, or **the header file** which contains the prototype.

This! This is a feature I would welcome into irony, to find the file pair to a file. We already have the compile options and what is left to do is to integrated with `ff-find-other-file`.

ikirill:
Why is this using `getCursorReferenced`? Shouldn't this be `getCursorDefinition`? [Its doc](http://clang.llvm.org/doxygen/group__CINDEX__CURSOR__XREF.html#gafcfbec461e561bf13f1e8540bbbd655b) says that it maps cursor to the definition cursor of an entity (rather than any of a number of its declarations). For a function this is a bit useless, but for a class, isn't this what you want?

Sarcasm:
It was first proposed to do this, see #152. I wasn't too fond of the gotchas that come with find-definition so find-declaration was proposed instead but I'm still not convinced enough.

xuhdev:
@Sarcasm If I add some Emacs integration, would you accept this PR, even as a temporary solution?

Sarcasm:
It depends how it is implemented.

We surely could benefit from knowing the prototype under the function at point, as a side not the location could be provided and you could jump to it if you want to.

I think I can accept something like this.

FWIW, what I was waiting for in terms of source code indexer seems to be out in the wild now, http://www.kythe.io/, but it still requires a lot of work to setup and to interface in a useful way with Emacs. Sourcegraph is another contender, https://github.com/sourcegraph/emacs-sourcegraph-mode, but doesn't support C/C++ for now...

wvxvw:
Since I was also interested in the feature, and as of right now won't be able to contribute code, but just a suggestion: when using SLIME and wanting to go to the definition of symbol at point, SLIME would create a "slide-in" buffer and display suggestions to what one could possibly mean by "definition".  Eg. say you point at a method.  This could have implicit or explicit generic function definition along with a couple of concrete method definitions all applicable to the same symbol.

So, "jump to definition" wouldn't take you immediately to any one of those locations, but offer a list of choices.  The ideal functionality for forward references and headers would be like that too: show the list of all things that may be relevant rather than try to guess.  (Now I know that this adds complexity to implementation, but at least this solves the "how to do it" part...)

PR #156, opened 2015-01-05
Add new directories to libclang header and lib search paths

YorkZ:
Added "/usr/local/llvm/VERSION/include..lib" to libclang header and lib search paths.


Sarcasm:
As said by email I don't think adding your custom path to irony-mode is the solution. Using CMAKE_PREFIX_PATH will be better.

Best,
Guillaume

Issue #138, opened 2014-12-07
std:: does not complete on Mac OS X


sebastiansturm:
Hi,

I have recently installed irony-mode on Mac OS X, with a homebrew-supplied libclang (from the 'llvm' package, currently at version 3.5). Unfortunately, the completion process doesn't pick up any standard library includes by default. For the following minimal example,

```
#include <string>
int main() {
  std::|
}
```

no completion candidates are offered. Running irony-server directly from the terminal shows that clang considers std an undeclared identifier. I can resolve the issue by explicitly passing the path to the stdlib header files to irony-server, i.e., by adding

```
   (list "-isystem" "/Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/include/c++/v1")
```

to the compile-options. Is there a better way to fix the stdlib completion issue? Since clang++ correctly picks up the header files during compilation (or flycheck syntax checking), I do not want to add the above include path to my compilation database.
best,
Sebastian

deniskokarev:
There is a good chance that you may don't even need brew (if you already have cmake).

irony-server compiles and works on OSX Yosemite with standard libclang from Xcode7. The only piece missing is clang-c header files, which may be borrowed from llvm 3.8 ;-)

Quick and dirty way to build irony-server on Yosemite with Xcode7
``` shell
### create /usr/local if not there yet
sudo mkdir -p /usr/local/lib /usr/local/include
### grab standard libclang
sudo cp -p "`xcode-select --print-path`"/Toolchains/XcodeDefault.xctoolchain/usr/lib/libclang.dylib /usr/local/lib
### (dirty part) supplement with latest llvm clang-c includes
cd /tmp
svn export http://llvm.org/svn/llvm-project/cfe/trunk/include/clang-c/
sudo cp -RP clang-c /usr/local/include
```

M-x irony-install-server should give you no trouble after that. Compiles and works.

One more step required on OSX - need to find where Xcode keeps all header files to add them to your .clang_complete. Otherwise irony is useless.

Peek at what Apple compiler uses
``` shell
echo | clang -x c++ -v -E - 2>&1 | sed -n '/^#include </,/^End/s|^[^/]*\([^ ]*/include[^ ]*\).*$|-I\1|p'
```
Add them to your .clang_complete:
```
$ cat .clang_complete
-I/Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/../include/c++/v1
[...]

PR #160, opened 2015-01-13
Made FindLibClang.cmake resiliant to homebrew llvm updates.

DavidTruby:
A recent homebrew llvm update broke the building of the server (since the llvm version changed to 3.5.0_2). However, I noticed that homebrew provides version independent paths to the llvm includes and libraries, so switching to these should fix not only this update but remove the need to update FindLibClang for similar homebrew llvm updates in the future.

### DIFF
--- a/server/cmake/modules/FindLibClang.cmake
+++ b/server/cmake/modules/FindLibClang.cmake
@@ -32,7 +32,7 @@ foreach (version ${LIBCLANG_KNOWN_LLVM_VERSIONS})
     # LLVM MacPorts
     "/opt/local/libexec/llvm-${version}/include"
     # LLVM Homebrew
-    "/usr/local/Cellar/llvm/${version}/include"
+    "/usr/local/opt/llvm/include"
     # LLVM Homebrew/versions
     "/usr/local/lib/llvm-${version}/include"
     )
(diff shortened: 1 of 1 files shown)


Sarcasm:
I actually would like to use `llvm-config[-VERSION]` instead of the default Homebrew prefixes, see https://github.com/Sarcasm/irony-mode/issues/138#issuecomment-65961146

Also I'm curious whether or not homebrew provides the CMake packages for LLVM, do you have `LLVMConfig.cmake` (not `LLVM-Config.cmake`) in /usr/local/Cellar/llvm35/3.5.0/lib/llvm-3.5?

What I have in mind is to look for `llvm-config` by default and potentially the user could specify the version to CMake (e.g: `cmake -DLLVM_CONFIG=llvm-config-3.5 <args...>`).

DavidTruby:
LLVMConfig.cmake is there, it's in the path /usr/local/Cellar/llvm/3.5.0_2/share/llvm/cmake/LLVMConfig.cmake. 

I think the issue with looking at llvm-config --version is that that reports 3.5.0, but the current package is actually unpacked in 3.5.0_2, so it wouldn't be possible to use that to look for the path it's unpacked at.

Homebrew also doesn't put llvm-config in the user's path so you still have the issue of having to look for that. I guess it might be possible to look at the paths returned by `brew list llvm`? I have no idea how complicated that will be.

Sarcasm:
Yeah the homebrew package is installed in 3.5.0_2, that's maybe what `llvm-config --prefix` returns. But what is the version of the clang directory present in the llvm install prefix `lib/clang/<VERSION>`. 
What's the output of `ls -Rd $(llvm-config --prefix)`?

DavidTruby:
Yeah `llvm-config --prefix` does return the correct path. The only issue is llvm-config not being put in the path by homebrew, but I guess you can look it up in `/usr/local/opt/llvm/bin/llvm-config`?

`ls -Rd $(/usr/local/opt/llvm/bin/llvm-config --prefix)` does return the correct path to llvm though. The clang path is `lib/clang/3.5.0` so it's not the same as the llvm one annoyingly.

PR #183, opened 2015-03-22
irony.el: check irony--server-process at irony-server-kill

simonbates:
Check that irony--server-process is non-nil before calling process-live-p.
Fixes error "Buffer ... has no process" when installing irony on a new
machine.

See https://github.com/Sarcasm/irony-mode/issues/145


Sarcasm:
Thanks, this should fix the duplicate issue #186.

PR #187, opened 2015-03-30
Irony-server: Detect libclang on FreeBSD.

josteink:
Without this fix getting irony-mode running is next to impossible on FreeBSD without intimiate knowledge of cmake build-flags, because building irony-server fails.

With this patch, it builds out of the box, creating a much better & portable experience.

### DIFF
--- a/server/cmake/modules/FindLibClang.cmake
+++ b/server/cmake/modules/FindLibClang.cmake
@@ -29,6 +29,7 @@ set(libclang_llvm_lib_search_paths
   )
 
 foreach (version ${LIBCLANG_KNOWN_LLVM_VERSIONS})
+  string(REPLACE "." "" undotted_version "${version}")
   list(APPEND libclang_llvm_header_search_paths
     # LLVM Debian/Ubuntu nightly packages: http://llvm.org/apt/
     "/usr/lib/llvm-${version}/include/"
@@ -38,6 +39,8 @@ foreach (version ${LIBCLANG_KNOWN_LLVM_VERSIONS})
     "/usr/local/Cellar/llvm/${version}/include"
     # LLVM Homebrew/versions
     "/usr/local/lib/llvm-${version}/include"
+    # FreeBSD ports versions
+    "/usr/local/llvm${undotted_version}/include"
     )
 
   list(APPEND libclang_llvm_lib_search_paths
@@ -49,6 +52,8 @@ foreach (version ${LIBCLANG_KNOWN_LLVM_VERSIONS})
     "/usr/local/Cellar/llvm/${version}/lib"
     # LLVM Homebrew/versions
     "/usr/local/lib/llvm-${version}/lib"
+    # FreeBSD ports versions
+    "/usr/local/llvm${undotted_version}/lib"
     )
 endforeach()
 


Sarcasm:
Thanks, but I'm not sure I want to add it. I believe my use of CMake is a bit unconventional for now and I don't want to add more of it.

Does FreeBSD exports the `LLVM-Config.cmake` file or installs the `llvm-config` binary?

I can add the undotted_version if needed but the port specific paths I'm not sure.

josteink:
I can sort of understand your mindset about this issue, but at the same time you've already special-cased for all other OSes and distros out there. Are you saying you will doubly special-case for FreeBSD and wont accept a patch which makes irony-mode work, unless that patch also solves how to resolve LLVM locations in general, across all platforms?

If so, I have to admit I find that a position bit disappointing.

To answer your questions: FreeBSD also supplies both a llvm-config binary and a LLVM-Config.cmake file. But these are (like the rest of llvm) stored in a versioned folder or accessed through a versioned binary available in the path:

This is for llvm35:

    $  find / -iname 'llvm-config*' 2>/dev/null
    /usr/local/share/doc/llvm35/html/llvm-config.html
    /usr/local/share/doc/llvm35/html/llvm-config.txt
    /usr/local/man/man1/llvm-config35.1.gz
    /usr/local/bin/llvm-config35 # exposed in path
    /usr/local/llvm35/bin/llvm-config
    /usr/local/llvm35/include/llvm/Config/llvm-config.h
    /usr/local/llvm35/share/llvm/cmake/LLVM-Config.cmake
    $

Other versions behave exactly the same, but with numbers replaced with other corresponding undotted versions.

Is that something you can work with? Do you see any other/better way to solve this without completely rewriting your cmake-file?

Sarcasm:
> I can sort of understand your mindset about this issue, but at the same time you've already special-cased for all other OSes and distros out there (even Windows!). Are you saying you will doubly special-case for FreeBSD by not accepting a patch which makes irony-mode work, unless that patch also solves how to resolve LLVM locations in general, across all platforms?

The reason I wasn't so keen to add FreeBSD is that the port path is user-configurable, it's not really like on Ubuntu where libclang is install to /usr/include and you are sure of it for the official package. This is an issue with the OS X package too, the Homebrew paths caused a lot of issues already.

> If so, I have to admit I find that a position bit disappointing.

I will accept you review, I guess you are right and what you propose is not worse than before.

> To answer your questions: FreeBSD also supplies both a llvm-config binary and a LLVM-Config.cmake file. But these are (like the rest of llvm) stored in a versioned folder or accessed through a versioned binary available in the path:

It's good to know that you have llvm-config and LLVM-Config.cmake.

At some point I would like to use `-DLLVM_CONFIG=llvm-config35` or something like that. So the user may have to specify it sometimes but the guess logic will be more straightforward.


> Is that something you can work with? Do you see any other/better way to solve this without completely rewriting your cmake-file?

Yeah, I actually want to rewrite this find_package logic but that's not the top priority, so adding a case for FreeBSD will do it for now.

Thanks for the patch!

Sarcasm:
Merged, thank you!

josteink:
Thanks. Glad you could approve it. FreeBSD needs some more love in the FOSS world :)

Sarcasm:
I used to use FreeBSD ;)

Issue #176, opened 2015-02-20
Reading large JSON databases is slow


Hylen:
Hi,

I've found that irony-cdb-autosetup-compile-options can be really slow when reading large JSON compilation databases. I noticed this when using irony-mode on llvm itself. Every time I open a new source file there is a ~5s delay.

Based on my attempts at profiling, it seems most of the time is spent in irony-cdb-json--transform-command. Time is also spent in json-read-file (emacs built-in). I am no elisp expert, but as I understand, this only reads and postprocess the database a little. The database file is ~10000 lines, 2 MB, so I don't understand where the 5s come form. Maybe it is the communication to the server? The elisp profiler may not see this time.

Anyway, I suggest moving the database reading to the server. I might also be willing to do this, if you agree. It would be a good way to read up libclang and one of my favorite emacs extensions :)

/Karl

Profiler output:
```
- ...                                                            3387  92%
 - irony-cdb--autodetect-compile-options                         2942  80%
  - catch                                                        2942  80%
   - let                                                         2942  80%
    - while                                                      2942  80%
     - let                                                       2942  80%
      - funcall                                                  2942  80%
       - irony-cdb-json                                          2942  80%
        - cond                                                   2942  80%
         - irony-cdb-json--get-compile-options                   2942  80%
          - let                                                  2942  80%
           - if                                                  2942  80%
            - progn                                              2942  80%
             - progn                                             2942  80%
              - let                                              2942  80%
               - irony-cdb-json--load-db                         2942  80%
                - delq                                           2942  80%
                 - mapcar                                        2939  80%
                  + irony-cdb-json--transform-compile-command               1958  53%
                  + json-read-file                                981  26%
   Automatic GC                                                   445  12%
+ command-execute                                                 254   6%
+ timer-event-handler                                              10   0%
```

Sarcasm:
Hi,

Yeah, I'm aware of the issue and I have some thoughts on that, I'm still not 100% sure about what's the proper solution.

> Based on my attempts at profiling, it seems most of the time is spent in irony-cdb-json--transform-command.

I'm not surprised here, one of the function that may be quite inefficient is `irony--split-command-line`.

> Time is also spent in json-read-file (emacs built-in). I am no elisp expert, but as I understand, this only reads and postprocess the database a little.

I think you understand right. The parsing method is not very efficient but I think it may be difficult to get one that scales well in Elisp (I may be wrong).

[...]

PR #188, opened 2015-04-03
libclang database

Hylen:
Broke out my changes to a new database, in response to #176.


irony-cdb-libclang.el
```diff
@@ -0,0 +1,89 @@
+;;; irony-cdb-libclang.el --- Compilation Database for irony using libclang
+
+;; Copyright (C) 2014  Guillaume Papin
```
Sarcasm: Put your name here ;)

irony-cdb-libclang.el
```diff
@@ -0,0 +1,89 @@
...
+;; MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+;; GNU General Public License for more details.
+
+;; You should have received a copy of the GNU General Public License
+;; along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+;;; Commentary:
+;;
+;; Compilation Database support for Irony using libclangs CXCompilationDatabase.
+;; Currently libclang only supports json compilation databases, but reading
```
Sarcasm: I don't know about this comment, since it is not entirely true maybe we will be better of without it.

The slowness comes from:

* json.el
* the fact that the Emacs built-in functions for splitting a command line are buggy
* that I had to roll my own parsing but it's badly implemented, very slow
* ...

irony-cdb-libclang.el
```diff
@@ -0,0 +1,89 @@
...
+;; using libclang is much faster than using json.el
+
+;;; Code:
+
+(require 'irony-cdb)
+(require 'cl-lib)
+(require 'pp)
```
Sarcasm: unused?

irony-cdb-libclang.el
```diff
@@ -0,0 +1,89 @@
...
+;;; Commentary:
+;;
+;; Compilation Database support for Irony using libclangs CXCompilationDatabase.
+;; Currently libclang only supports json compilation databases, but reading
+;; using libclang is much faster than using json.el
+
+;;; Code:
+
+(require 'irony-cdb)
+(require 'cl-lib)
```
Sarcasm: nitpicking:

I add a newline between irony's own `require` "local requires" and "third-party requires", and built-in requires, like I do in C++ with includes.

So I would favor:

```el
(require 'irony-cdb)

(require 'cl-lib)
```

irony-cdb-libclang.el
```diff
@@ -0,0 +1,89 @@
...
+;; but WITHOUT ANY WARRANTY; without even the implied warranty of
+;; MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+;; GNU General Public License for more details.
+
+;; You should have received a copy of the GNU General Public License
+;; along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+;;; Commentary:
+;;
+;; Compilation Database support for Irony using libclangs CXCompilationDatabase.
```
Sarcasm: Maybe add a link to http://clang.llvm.org/doxygen/group__COMPILATIONDB.html?

irony-cdb-libclang.el
```diff
@@ -0,0 +1,89 @@
...
+
+(defvar-local irony-cdb-libclang--server-compile-options nil
+  "Compiler options got from irony-server.")
+
+(defun irony-cdb-libclang--request-handler (flags result)
+  (set result flags))
+
+;; TODO: sometimes there is garbage on the callback-queue, determine why
+(defun irony-cdb-libclang--get-compile-options ()
+  (irony--awhen (irony-cdb-json--locate-db)
```
Sarcasm: irony-cdb-json isn't required at the top file but it is used here.

I think irony-cdb should offer some kind of build-directory support. I guess for now we can keep the logic as-is but I would prefer `irony-cdb-json--locate-build-directory` for a first step, instead of calling `file-name-directory` in `irony-cdb-libclang--server-exact-flags`.

irony-cdb-libclang.el
```diff
@@ -0,0 +1,89 @@
...
+    (let ((db-location it))
+      ;; Try to obtain flags from server
+      (irony--aif (irony-cdb-libclang--server-exact-flags db-location)
+          it
+        ;; Couldn't get flags from server. Load db using irony-cdb-json guess
```
Sarcasm: No, I think you should just return `nil` instead, it means that irony-cdb will call the next compilation database, `irony-cdb-json`.

irony-cdb-libclang.el
```diff
@@ -0,0 +1,89 @@
...
+        ;; the flags
+        (let* ((db (irony-cdb-json--load-db db-location))
+               (dir-cdb (irony-cdb-json--compute-directory-cdb db)))
+          (irony-cdb-json--guess-flags dir-cdb))))))
```
Sarcasm: In the end this function should be as simple as:

```el
(irony--aif (irony-cdb-json--locate-build-directory)
  (irony-cdb-libclang--server-exact-flags it))
```

irony-cdb-libclang.el
```diff
@@ -0,0 +1,89 @@
...
+
+(defun irony-cdb-libclang--server-exact-flags (db-file)
+  "Get compile options from server"
+  (let ((project-root (file-name-directory db-file))
+          (file buffer-file-truename))
```
Sarcasm: Careful, what if you have no file attached to this buffer?
`buffer-file-truename` returns the abbreviated filename, i.e: `~/`, I don't think that's what we want.

`file` is used for two things:

1. sending as argument to irony-server's get-compile-options, here we don't want to use `buffer-file-truename` but `buffer-file-name`. Tthis one doesn't exhibit the abbreviated-filename behavior, and the CXCompilationDatabase has the logic to handle truename if that is needed.
2. `irony-cdb-libclang--adjust-options-and-remove-compiler` that forwards it to `irony-cdb-json--adjust-compile-options`. Here we also want the original filename, `irony-cdb-json--adjust-compile-options` will should be responsible of calling truename if need be.

See `irony--get-buffer-path-for-server` logic too.

irony-cdb-libclang.el
```diff
@@ -0,0 +1,89 @@
...
+
+(defun irony-cdb-libclang--adjust-options-and-remove-compiler (file cmds)
+  "Remove compiler, target file and output file from cmds
+
+cmds is a list of conses, with car holding the options and cdr holding the
+working directory where the command was issued."
+  (mapcar (lambda (cmd)
+          (let ((opt (car cmd))
+                (wdir (cdr cmd)))
+            (cons (cdr
```
Sarcasm: This `cdr` removes the compilers? 

Removing the compiler should be done before the call to `irony-cdb-json--adjust-compile-options`, this function expects compile options as argument.

Maybe make the removing of the compiler needs to be quite explicit (i.e: a function of it's own), so that we can provide it in `irony-cdb`. I think in the future we may need to handle some things like compiler-wrappers, e.g: ccache.  A dedicated function could be useful.

irony.el
```diff
@@ -661,6 +661,17 @@ If no such file exists on the filesystem the special file '-' is
...
 
+(defun irony--send-request-sync (request callback &rest args)
+  "Send a request to irony-server and wait for the result.
+
+callback should set the irony-sync-result property of the irony-server process"
+  (let ((process (irony--get-server-process-create)))
+    ;; Using apply to "spread" the args list
+    (apply 'irony--send-request request callback args)
+    ;; While there are callbacks left on the queue, wait for output from server
+    (while (process-get process 'irony-callback-stack)
```
Sarcasm: You should make a function of it's own for the check of the callback stack, `irony--server-process-has-callback-p` or something along these lines, it's not impossible that I rework the callback stack mechanism a little bit, that would make it easier for me if it's already along with the other callback-stack manipulation functions.

irony.el
```diff
@@ -661,6 +661,17 @@ If no such file exists on the filesystem the special file '-' is
                              (format "%s\n"
                                      (combine-and-quote-strings argv)))))))
 
+(defun irony--send-request-sync (request callback &rest args)
+  "Send a request to irony-server and wait for the result.
+
+callback should set the irony-sync-result property of the irony-server process"
+  (let ((process (irony--get-server-process-create)))
+    ;; Using apply to "spread" the args list
```
Sarcasm: That the use of `apply`, not sure if this comments adds much.

irony.el
```diff
@@ -661,6 +661,17 @@ If no such file exists on the filesystem the special file '-' is
                              (format "%s\n"
                                      (combine-and-quote-strings argv)))))))
 
+(defun irony--send-request-sync (request callback &rest args)
+  "Send a request to irony-server and wait for the result.
+
+callback should set the irony-sync-result property of the irony-server process"
```
Sarcasm: What is `irony-sync-result`?

irony.el
```diff
@@ -661,6 +661,17 @@ If no such file exists on the filesystem the special file '-' is
...
+  (let ((process (irony--get-server-process-create)))
+    ;; Using apply to "spread" the args list
+    (apply 'irony--send-request request callback args)
+    ;; While there are callbacks left on the queue, wait for output from server
+    (while (process-get process 'irony-callback-stack)
+      (accept-process-output process))))
```
Sarcasm: You need to handle some kind of local quit. Quoting `4.` from https://github.com/Sarcasm/irony-mode/issues/176#issuecomment-77773326

> when using accept-process-output, C-g should be allowed to not freeze Emacs if something goes wrong, see https://www.gnu.org/software/emacs/manual/html_node/elisp/Timers.html. FYI, a long time ago, irony used to be blocking and I was using this function to wait for server response: https://github.com/Sarcasm/irony-mode/blob/0111357aff8c30a7fa54393365e822274f0ca739/elisp/irony.el#L371

Notably the function checks whether or not the process died, I'm not sure if that's necessary but that may be.

irony.el
```diff
@@ -661,6 +661,17 @@ If no such file exists on the filesystem the special file '-' is
                              (format "%s\n"
                                      (combine-and-quote-strings argv)))))))
 
+(defun irony--send-request-sync (request callback &rest args)
```
Sarcasm: Thinking about it, the logic isn't quite right here, the check of the callback stack doesn't make so much sense, you should probably register a callback that sets some kind of flag, when the flags is set it means the response has been received.

server/src/Command.cpp
```diff
@@ -99,6 +99,7 @@ std::ostream &operator<<(std::ostream &os, const Command::Action &action) {
 std::ostream &operator<<(std::ostream &os, const Command &command) {
   os << "Command{action=" << command.action << ", "
      << "file='" << command.file << "', "
+     << "dir='" << command.dir << "', "
```
Sarcasm: The only use we have for now is the build-dir? I prefer the use of `buildDirectory` for now.

server/src/Command.h
```diff
@@ -43,6 +43,7 @@ struct Command {
 
   std::vector<std::string> flags;
   std::string file;
+  std::string dir;
```
Sarcasm: buildDirectory

server/src/Commands.def
```diff
@@ -18,7 +18,8 @@ X(Exit, "exit", "exit interactive mode, print nothing")
 X(Help, "help", "show this message")
 X(Parse, "parse", "FILE - parse the given file")
 X(SetDebug, "set-debug", "[on|off] - enable or disable verbose logging")
-
+X(GetCompileOptions, "get-compile-options", "PROJECT_ROOT FILE - "
```
Sarcasm: PROJECT_ROOT -> BUILD_DIRECTORY

server/src/CompilationDatabase.cpp
```diff
@@ -0,0 +1,75 @@
+/**
+ * \file
+ *
+ * \brief Using libclang for reading the JSON compilation database.
+ *
+ * This file is distributed under the GNU General Public License. See
+ * COPYING for details.
+ */
+#include <iostream>
```
Sarcasm: Order of includes is reversed here, I go from the most specific to the less specific.

See http://llvm.org/docs/CodingStandards.html#include-style

server/src/CompilationDatabase.cpp
```diff
@@ -0,0 +1,75 @@
+/**
```
Sarcasm: The function in this file should directly go in Irony.cpp.

server/src/CompilationDatabase.cpp
```diff
@@ -0,0 +1,75 @@
...
+ */
+#include <iostream>
+
+#include <clang-c/Index.h>
+#include <clang-c/CXCompilationDatabase.h>
+#include <clang-c/CXString.h>
+
+#include "CompilationDatabase.h"
+
+static void getExactFlags(std::vector<std::vector<std::string>> &flags,
```
Sarcasm: Out parameters goes last. But here it shouldn't be so much of a performance bottleneck to return by value, I think it will be more natural to do so.

server/src/CompilationDatabase.cpp
```diff
@@ -0,0 +1,75 @@
...
+                          CXCompilationDatabase &db,
+                          const std::string &fullFilename) {
+  unsigned ncmd;
```
Sarcasm: In C++, unlike C we can create variables wherever we sit fit. It's better for code locality to declare them when used directly, at the smallest scope necessary.

So it should read:

```cpp
CXCompileCommands cmds = clang_CompilationDatabase_getCompileCommands(db, fullFilename.c_str());

for (unsigned i = 0, ncmd = clang_CompileCommands_getSize(cmds); i < ncmd; ++i) {
   // same inside the for-loop...
}
```

server/src/CompilationDatabase.cpp
```diff
@@ -0,0 +1,75 @@
...
+std::vector<std::vector<std::string>>
+getFlags(const std::string &projectRoot, const std::string &fullFilename) {
+  std::vector<std::vector<std::string>> flags;
+  CXCompilationDatabase db;
+  CXCompilationDatabase_Error error;
+
+  db = clang_CompilationDatabase_fromDirectory(projectRoot.c_str(), &error);
+
+  if (error == CXCompilationDatabase_CanNotLoadDatabase) {
+    std::clog << "Cannot load database!\n";
```
Sarcasm: You should log the database maybe so the log is more helpful.

Right now the code isn't super consistent, but for log I prefer errors, warnings, ...with some kind of markers.

Something along the line of:

```cpp
    std::clog << "W: failed to load database from " << buildDirectory << std::endl;
```

server/src/Irony.cpp
```diff
@@ -326,4 +327,34 @@ void Irony::complete(const std::string &file,
     clang_disposeCodeCompleteResults(completions);
     std::cout << ")\n";
   }
+
+}
+
+// Get compile options from JSON database
+void Irony::getCompileOptions(const std::string &projectRoot,
```
Sarcasm: Please use buildDirectory over projectRoot.

server/src/CompilationDatabase.cpp
```diff
@@ -0,0 +1,75 @@
...
+    unsigned j;
+    CXCompileCommand cmd;
+    CXString cxdir;
+    const char *dir;
+
+    cmd = clang_CompileCommands_getCommand(cmds, 0);
+    cxdir = clang_CompileCommand_getDirectory(cmd);
+
+    dir = clang_getCString(cxdir);
+    flags[i].push_back(dir);
```
Sarcasm: why or you pushing dir here, that's a little bit weird.

server/src/Irony.cpp
```diff
@@ -326,4 +327,34 @@ void Irony::complete(const std::string &file,
...
+  std::vector<std::vector<std::string>> flags = getFlags(projectRoot, file);
+
+  if (!flags.empty()) {
+    std::cout << "(\n";
+
+    for (auto vi = flags.begin(), vend = flags.end(); vi != vend; ++vi) {
+      std::cout << "( ";
+
+      std::cout << "(";
+      for (auto it = std::next(vi->begin()), end = vi->end(); it != end; ++it)
```
Sarcasm: The build-directory being first element of flags is weird, we may be better off with a pair.

server/src/Irony.h
```diff
@@ -78,6 +78,16 @@ class Irony {
                 const std::vector<CXUnsavedFile> &unsavedFiles);
   /// \}
 
+  /// \brief Get compile options from JSON database.
+  ///
+  /// \param projectRoot Directory containing compile_commands.json
```
Sarcasm: \param buildDirectory The project build directory.

server/src/Irony.h
```diff
@@ -78,6 +78,16 @@ class Irony {
...
+  /// \param file File to obtain compile commands for.
+  ///
+  /// \todo Return all compile commands for a file, not just the first.
```
Sarcasm: It's done, isn't it?

irony.el
Hylen: My thought was that once a synchronous request has been made, all previous requests must finish. I guess that I missed that callbacks can themselves register new requests? Or are you thinking about something else?

Sarcasm: Yeah that was what I was thinking about.

I was also thinking about that this way is a little bit more straightforward, and less risky if there are some other factors that I haven't foreseen.

Sarcasm:
Closing in favor of #208.

PR #189, opened 2015-04-08
Fixed bug in irony--parse-request-handler.

Hylen:
irony--parse-request-handler should use the parsed buffer, not the current. Probably the cause of Sarcasm/flycheck-irony#3.

### DIFF
--- a/irony.el
+++ b/irony.el
@@ -718,20 +718,21 @@ care of."
   (equal irony--parse-buffer-state
          (cons (irony--parse-buffer-context) 'requested)))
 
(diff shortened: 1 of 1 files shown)


Sarcasm:
There was an indentation issue, I fixed it before committing. The commit is https://github.com/Sarcasm/irony-mode/commit/17763686848e81fbee13debdfc388f017dca52af (github doesn't link it to the PR apparently).

Thanks for the fix, I had to do another fix in irony-diagnostics.el for flycheck-irony to be totally fixed.

Issue #200, opened 2015-06-08
Not including private fields possible?


dzikakulka:
Would it be posible to make Irony parser or whatever magic it uses to list my variables ignore private fields in classes? I get that it wouldn't be all-aroung great option since people friend stuff and they may want them available then (and I'm new to programming so there may be instances where this could be needed as well), but if someone could hint me if that was at all possible and what needs to be modified, I would be very grateful.

Cheers!

Sarcasm:
It's on my todo list, I make very slow progress theses days however. It is possible with the libclang API, I haven't looked seriously yet.

dzikakulka:
Great! It's very nice to see active devs around emacs tools (even if progress goes slowly) - thanks for making my environment nice and smooth. Cheers!

Sarcasm:
To whoever want to give it a try, I think there is something to do with the function [clang_getCompletionAvailability()](http://clang.llvm.org/doxygen/group__CINDEX__CODE__COMPLET.html#gacbbded5dd9c27b927ed5080d8b530845).

PR #203, opened 2015-06-14
Filter unavailable and inaccessible completions

aport:
Based on the comments in https://github.com/Sarcasm/irony-mode/issues/200 I wanted to give this is shot. It was a lot more straightforward than I anticipated.

This skips completions that are unavailable or inaccessible and would result in error if used.

I tested this out with some classes containing private members and verified that those members no longer showed up in the company-mode completion list.


Sarcasm:
Great, thank you, I knew it was probably easy. ;)

PR #207, opened 2015-06-22
Expand paths for JSON CDB when read interactively.

cahlbin:
This will write the project root and CDB file path as absolute paths when written to cdb-json-projects


Sarcasm:
I merged #216 instead which is slightly more correct I think, in case the function isn't called interactively.

Thanks.

PR #212, opened 2015-07-08
Added a compilation database based on libclang.

Hylen:


Sarcasm:
Merged manually (rebased + commit message amended) in 9ae5381fdd0c1f0e5378c6fe9bf757630dc61e13.

Thank you very much Hylen, I really appreciate this feature, this is the most useful feature that came in recently. I still have no idea what is the proper way to handle compile options for headers.

Hylen:
No problem, I enjoy using Irony-mode and it was fun contributing.

Sarcasm:
Hello @Hylen,

I noticed a small bug today and I will correct it soon enough but I have a question first.

When I open a buffer, my irony-mode hooks are called twice.
Could you tell me if it happens to you to?

```el
(defun irony-mode-hello-hook ()
  (message "irony-mode-hook called"))

(add-hook 'irony-mode-hook 'irony-mode-hello-hook)
```

Actually, I have this issue seems to be with `c-mode-hook`, `c++-mode-hook`, even in `emacs -Q` for me.

This bug is annoying, because the get-compile-options is already quite slow (the elisp version, not yours) and now it is called twice, so that doesn't either...

Sarcasm:
For reference, the commit that fixes the issue has been pushed, see 6d3d46eefbf49fc0ef1e0beb62b2bd33cb7d37ae

Hylen:
Nice catch on noticing the bug! I also have the hook called twice, and see no obvious explanation. It is probably a bug in c++-mode.

PR #216, opened 2015-07-17
Can now use relative paths to project root and compile commands.

Hylen:
Previously, relative paths to project root and compile commands caused problems. If for instance project root was specified using ~, irony-cdb-json--find-best-prefix-path would fail to match.


Sarcasm:
Thanks.

PR #229, opened 2015-09-07
Get type of a symbol

hotpxl:
Hi.

As we discussed last time, the getting type of symbol bit is most useful and easiest to implement at current stage. I've modified the server a bit to include this functionality from Clang. It works on simple examples, the command line arguments are just like those of auto-completion. Due to some Clang restrictions, it is not always able to deduce what `auto` means, particularly in the definition of the variable. But it works like a charm in other cases. I pulled most of the syntax from YouCompleteMe. I haven't programmed using Clang before so I'm not sure about memory leaks and such.

This is just a demo of idea and might contain bugs or unhandled situations. What do you think about this integration?


Sarcasm:
Interestingly, `auto` seems to have been fixed this week: https://llvm.org/bugs/show_bug.cgi?id=18669 / http://reviews.llvm.org/D11976

I haven't been able to test yet.

Sarcasm:
Yep `auto` seems to work on top-of-trunk, thanks for the snippet, will be nice to integrate it.

hotpxl:
Great. I'll make the server side more stable and try to implement the elisp part.

hotpxl:
Currently I test with files on my disk. What would happen if I feed an unsaved file from editor? What precautions do I have to pay?

Sarcasm:
You do not have to pay much attention.
Actually I think "get-type" should be implemented the same way as "diagnostics", a precondition is to have a successful parse preceding it.

hotpxl:
Good. I just wonder how unsaved buffers are handled.

I don't see it implemented as "diagnostics", since "get-type" should take a cursor location. I think it should be look like "complete".

Sarcasm:
Unsaved buffer will be handled by parse.

hotpxl:
Hi. I've added the `irony-get-type` function and it's working like it is in YCM now.

How's your opinion on this?

hotpxl:
Hi. This has been a long time. Do you have any plan of integrating this feature? Thanks!

Sarcasm:
Hello, sorry for the enourmous delay :)

I would change one thing like I said earlier, this should be implemented the same way as diagnostics, that is, a previous "parse" should be required. You should "copy/paste" irony-diagnostics-async, that should do it for now.

I can do it later in the week if you want.

Sorry for not considering this issue more seriously, I'm pretty sure I will appreciate using it.

Thank you.

hotpxl:
No need to apologize. Getting DOSed by todo list happens to everyone. It would be great if you could finish this off. I'm not sure if I get what you mean.

hotpxl:
No need to apologize. Getting DOSed by todo list happens to everyone. It would be great if you could finish this off. I'm not sure if I get what you mean.

Sarcasm:
I committed this as https://github.com/Sarcasm/irony-mode/commit/cd064803d4e90d667b8c75e4c86e76e0cde21176

Let me know if it still works for you.
What binding do you use? Maybe the README should be updated with some recommended binding.

Best,
Guillaume

hotpxl:
Awesome. It works perfectly. Personally I use `,t` with `evil-leader`. It comes from YCM under vim.

I would recommend using `C-c i` as prefix for irony-mode, so `get-type` would be mapped to `C-c i t`. What do you think about it?

Sarcasm:
> I would recommend using C-c i as prefix for irony-mode, so get-type would be mapped to C-c i t. What do you think about it?
>  Hide all checks

Agreed. I'm have to look how to make a prefix map. :)

dominikh:
> I would recommend using C-c i as prefix for irony-mode

Please be aware that `C-c <single letter>` is reserved for users and shouldn't be used by modes: http://www.gnu.org/software/emacs/manual/html_node/elisp/Key-Binding-Conventions.html

Issue #236, opened 2015-10-08
irony-cdb-json-add-compile-commands-path doesn't work


MikeSanMichel2:
#234 was about finding `build/compile_commands.json` automatically, but
after more testing I think irony doesn't actually use a cdb if autosetup can't find it.

Using `irony-cdb-json-add-compile-commands-path`  to add  `build/compile_commands.json` doesn't work. it updates `irony-cdb-json--project-alist`, and it shows up in `~/.emacs.d/irony/cdb-json-projects` but the completion doesn't show members, and flycheck-irony complains about headers (missing `-I`).

Irony provides some completion even without  a cdb, which makes it seem like it's
working, but so far the only thing that actually works is putting a symlink to `compile_commands.json`
in the project root and running `autosetup`. a few seconds after that the flycheck errors
go away and completion finds class members.

`cpputils-cmake` makes no difference. if the cdb is in the root irony works, otherwise it doesn't.

`irony-20150831.144/`

MikeSanMichel2:
Looks like #193 reported same.
And #190 too.

MikeSanMichel2:
Looks like I was fooled by the UI, and used the build directory as project root.
So not a bug, but my mistake. Will follow up with a PR to perhaps simplify things
and eliminate the chance of error.

Sarcasm:
Thanks ;)

PR #237, opened 2015-10-08
Useful additions to irony-cdb-json

### LINKED ISSUE #234: Request: better support for cmake out-of-tree build directories
Hi,

Irony has been great, but I had some trouble getting it to work. Here's one
suggestion/request for a better out-of-the-box experience.

- Many popular OSS projects use cmake and I use it for my own as well.
- The norm in CMake is that you use a seperate build directory instead of building in-tree.
- CMake will generate `compile_commands.json`, if you enable `CMAKE_EXPORT_COMPILE_COMMANDS` on the commandline or in `CMakeLists.txt`.
- The cdb ends up in the build directory.
- Often (though not always) the build directory is named `build` under the project's root
directory.

Irony currently only searches parent directories (recursively) for the cdb, so I've had to
place a symlink in the project root. Since CMake is so common and there's a convention
on where the build directory lives, would it be possible to have irony check for `build/compile_commands.json` as it traverses through parent directories? that would
eliminate one step in getting irony setup.

Thanks again for making the package available!

MikeSanMichel2:
closes #234, #236 maybe some issues linked there as well. 

First, now looks recursively for parents for both ".json" and "build/.json". (#234)

Second, #236. One confusing bit is that calling `irony-cdb-json-add-compile-commands-path`
doesn't actually finish by loading the database you added. That might be "Cleaner", but
its inconvenient and under-documented. Commit 2 fixes that.

Third, still #236, I was led astray by `add-compile-commands-path` asking for both
project root and cdb path and ended up choosing the build directory, then the file within it.
I may not have been paying enough attention, but the issues suggest I may not be
the only one. So, commit 3 adds an alternative which takes only a cdb path,
loads it, and uses the source file paths in it to find the project root as the deepest
dir parenting everything in the cdb.
I would have just replaced `add-compile-commands-path` entirely, but that's public
and the new function takes one argument less. If you like the idea and don't mind
breaking the API, go for it.

Please review/test, I only wrote it, I haven't actually tested it (much) :)

If you merge 1m my problem is basically solved. 2 I think is reasonable. 3 was just
frustration.

### DIFF
--- a/irony-cdb-json.el
+++ b/irony-cdb-json.el
@@ -66,7 +66,10 @@ directories to project directory."
   (add-to-list 'irony-cdb-json--project-alist
                (cons (expand-file-name project-root)
                      (expand-file-name compile-commands-path)))
-  (irony-cdb-json--save-project-alist))
+  (irony-cdb-json--save-project-alist)
+
+  ; and tell irony to load it now
+  (irony-cdb-autosetup-compile-options))
 
 (defun irony-cdb-json--get-compile-options ()
   (irony--awhen (irony-cdb-json--locate-db)
--- a/irony-cdb-json.el
+++ b/irony-cdb-json.el
@@ -115,9 +115,11 @@ directories to project directory."
 
 (defun irony-cdb-json--locate-db ()
   (irony-cdb-json--ensure-project-alist-loaded)
-  (irony--aif (locate-dominating-file (irony-cdb-json--target-path)
-                                      "compile_commands.json")
-      (expand-file-name "compile_commands.json" it)
+  (irony--aif (irony-cdb--locate-dominating-file-with-dirs
+               (irony-cdb-json--target-path)
+               "compile_commands.json"
+               irony-cdb-search-directory-list)
+      (expand-file-name it)
     ;; if not in a parent directory, look in the project alist
     (irony--awhen (irony-cdb-json--find-best-prefix-path
                    (irony-cdb-json--target-path)
--- a/irony-cdb.el
+++ b/irony-cdb.el
@@ -63,6 +63,20 @@ for files that it cannot handle."
   :type '(repeat function)
   :group 'irony-cdb)
 
+(defcustom irony-cdb-search-directory-list '("." "build")
+  "List of relative subdirectory paths to be searched for cdb files
+
+Irony looks for cdb files in any of the supported format by checking
+each directory from the currently loaded file and recursively through
+parent directories until it hits the root directory or a cdb is
+found. At each level of the search irony looks at the subdirectories
+listed in `'irony-cdb-search-directory-list` for the files. Customize this
+list if your cdb is held in a custom directory within you project,
+such as a custom named build directory.
+"
+  :type '(repeat string)
+  :group 'irony-cdb)
+
 
 ;;
 ;; Internal variables
--- a/irony-cdb.el
+++ b/irony-cdb.el
@@ -128,6 +128,57 @@ such as a custom named build directory.
 ;; Functions
 ;;
 
+(defun irony-cdb--choose-closest-path (file paths)
+  "Find the \"best\" path in PATHS matching FILE
+
+If any paths in PATHS is belongs to the same directory
+or a subdirectory of file, we disregard other candidates.
+
+For remaining candidates, \"nearest\" is measured as abs. difference
+in path depth.
+- We prefer deeper paths at level +N to those at level -N.
+- If multiple paths are equally good, we return the last one.
+
+Returns nil if paths isn't a list of at least one element.
+"
+  (when (listp paths)
+    (let ((paths (or
+                  ;; if we find a cdb in cwd or below, don't consider other candidates
+                  (cl-remove-if-not (lambda (x) (string-prefix-p (file-name-directory file) x)) paths)
+                  paths)))
+      (cl-loop for path in paths
+         with best-depth-delta = 999999 ; start at +inf
+         with best-path = nil ; we keep the best so far here
+         ;; all candidates have their depth compared to that of target file
+         with file-depth = (length (split-string file "/")) ;
+         for candidate-depth = (length (split-string path "/"))
+         ;; Our metric. We use signum as a tie-breaker to choose deeper candidates
+         for depth-delta = (+ (abs (- file-depth candidate-depth))
+                              (* 0.1 (- file-depth candidate-depth)))
+         do (when (< depth-delta best-depth-delta)
+              (progn
+                (setq best-depth-delta depth-delta)
+                (setq best-path path)))
+         finally return best-path))))
+
+(defun irony-cdb--locate-dominating-file-with-dirs (file
+                                                    name
+                                                    subdirectories)
+  "Convenience wrapper around `locate-dominating-file'
+
+Looks up the directory hierarchy from FILE for to locate any directory
+in `subdirectories` which contains NAME. If multiple files are found,
+chooses the one located at the nearest directory level. if multiple
+files are found at the same level, picks the first one encountered.
+returns the full path to file if found, or nil otherwise."
+  (let ((candidates
+         (cl-loop for subdir in subdirectories
+            for relpath = (concat (file-name-as-directory subdir) name)
+            for match-maybe = (locate-dominating-file file relpath)
+            when match-maybe collect (expand-file-name (concat match-maybe relpath)))))
+    (irony-cdb--choose-closest-path file candidates)))
+
+
 (defun irony-cdb--update-compile-options (compile-options
                                           &optional working-directory)
   (setq irony--compile-options compile-options
--- a/server/test/elisp/irony-cdb-json.el
+++ b/server/test/elisp/irony-cdb-json.el
@@ -4,6 +4,7 @@
          "test-config"))
 
 (require 'irony-cdb-json)
+(require 'cl-lib)
 
 (defconst irony-cdb/compile-command
   '((file      . "../src/file.cc")
@@ -57,3 +58,30 @@
    (equal nil
           (irony-cdb--choose-closest-path "/tmp/a/1"
                                           'ordures))))
+
+; http://endlessparentheses.com/understanding-letf-and-how-it-replaces-flet.html
+(ert-deftest cdb/locate-db/choose-among-candidates ()
+  (should
+   (equal "/foo/build/cdb"
+          (cl-letf (((symbol-function 'locate-dominating-file)
+                     (lambda  (file name)
+                       (cond
+                         ((string= name "./cdb") "/") ; found /cdb
+                         ((string= name "build/cdb") "/foo/") ; found /foo/build/cdb
+                         ))))
+            (irony-cdb--locate-dominating-file-with-dirs "/foo/bar/qux.cpp"
+                                                         "cdb"
+                                                         '("." "build" "out/x86_64"))))))
+
+(ert-deftest cdb/locate-dominating-file-with-dirs/children-first ()
+  (should
+   (equal "/tmp/foo/bar/out/x86_64/cdb"
+          (cl-letf (((symbol-function 'locate-dominating-file)
+                     (lambda  (file name)
+                       (cond
+                         ((string= name "./cdb") "/tmp/foo/") ; found /tmp/foo/cdb
+                         ((string= name "out/x86_64/cdb") "/tmp/foo/bar/") ;found /tmp/foo/bar/out/x86_64/cdb
+                         ))))
+                    (irony-cdb--locate-dominating-file-with-dirs "/tmp/foo/bar/qux.cpp"
+                                                                 "cdb"
+                                                                 '("." "out/x86_64" ))))))
--- a/server/test/elisp/irony-cdb-json.el
+++ b/server/test/elisp/irony-cdb-json.el
@@ -27,3 +27,33 @@
    (equal "/home/user/project/build"
           (nth 2 (irony-cdb-json--transform-compile-command
                   irony-cdb/compile-command)))))
+
+(ert-deftest cdb/choose-closest-path/chooses-closest ()
+  (should
+   (equal "/tmp/a/cdb"
+          (irony-cdb--choose-closest-path "/tmp/a/1"
+                                          '("/tmp/a/cdb" "/tmp/cdb")))))
+
+(ert-deftest cdb/choose-closest-path/chooses-closest2 ()
+  (should
+   (equal "/tmp/a/cdb"
+          (irony-cdb--choose-closest-path "/tmp/a/1"
+                                          '("/tmp/cdb" "/tmp/a/cdb")))))
+
+(ert-deftest cdb/choose-closest-path/prefers-deeper ()
+  (should
+   (equal "/tmp/a/build/cdb"
+          (irony-cdb--choose-closest-path "/tmp/a/1"
+                                          '("/tmp/a/build/cdb" "/tmp/cdb")))))
+
+(ert-deftest cdb/choose-closest-path/prefers-deeper2 ()
+  (should
+   (equal "/tmp/a/build/cdb"
+          (irony-cdb--choose-closest-path "/tmp/a/1"
+                                          '("/tmp/cdb" "/tmp/a/build/cdb")))))
+
+(ert-deftest cdb/choose-closest-path/will-survive-garbage ()
+  (should
+   (equal nil
+          (irony-cdb--choose-closest-path "/tmp/a/1"
+                                          'ordures))))


MikeSanMichel2:
Nice. appveyor seems unrelatedly broken.

Sarcasm:
> Nice. appveyor seems unrelatedly broken.

Do not worry about appveyor for now, I tried to disable it.

There is some nices ideas in this PR, I will have to take a deeper look during the week if possible, do not hesitate to remind me. ;)

irony-cdb-json.el
```diff
@@ -111,17 +161,19 @@ directories to project directory."
            finally return found))
 
 (defun irony-cdb-json--locate-db ()
```
Sarcasm: 1. Do I understand correctly that the `irony-cdb-json--project-alist` is no longer used?
2. What about people who put their build directory in `Build`, or maybe, `build/release`, `build/Debug`, `build-clang`, `build-gcc-x86_64`, ...? I don't really want to hardcode `build` anywhere in irony, it should just be easy for the user to customize the lookup with his preferred builds.
3. please favor `locate-dominating-file` over custom logic. To understand why, see `locate-dominating-stop-dir-regexp`, `locate-dominating-file`, `expand-file-name`, ...

irony-cdb-json.el
```diff
@@ -48,6 +48,53 @@ source directory.")
...
+         (minlen (apply 'min (mapcar 'length dir-components)))
+         (common-count (cl-loop for i in (number-sequence 0 minlen)
+                                while (= 1 (length (delete-dups (mapcar (lambda (val) (nth i val)) dir-components))))
+                                finally return i
+                                )))
+    (mapconcat 'file-name-as-directory (cons "/" (-take common-count (car dir-components))) "")
+    ))
+
+;;;###autoload
+(defun irony-cdb-json-add-compile-commands-path-auto-root (compile-commands-path)
```
Sarcasm: I would change `irony-cdb-json-add-compile-commands-path` to take the arguments in the reverse order.
First the build directory or compilation database.
Then the second argument could be pre-filled with what you guessed with `irony-cdb-json--find-common-root-directory`.
We will have only one function but the user is helped a bit so he can just press `RET` but he is still able to specify a custom build dir (the guess will be wrong on quite a few of my projects).

irony-cdb-json.el
```diff
@@ -66,7 +113,10 @@ directories to project directory."
   (add-to-list 'irony-cdb-json--project-alist
                (cons (expand-file-name project-root)
                      (expand-file-name compile-commands-path)))
-  (irony-cdb-json--save-project-alist))
+  (irony-cdb-json--save-project-alist)
+
+  ; and tell irony to load it now
+  (irony-cdb-autosetup-compile-options))
```
Sarcasm: This is simple but useful, thanks.

Sarcasm:
I'm still thinking about the compilation database stuff.
I am wondering if the compilation database should provide a command to guess/find the build directory.
So maybe the logic of finding the build directory can be deported to compilation database somehow (that's not clear to me).

irony-cdb-json.el
MikeSanMichel2: Will fix.

irony-cdb-json.el
MikeSanMichel2: Though about that, but what about uses who have calls setup with explicit arguments in their .emacs or some hook? If you don't care, I'll gladly make this change.

MikeSanMichel2:
What does irony need the build directory for? I've only just looked at this tiny piece of the package.

irony-cdb-json.el
Sarcasm: I'm okay with changing this function, compilation database stuff is not stable.

Sarcasm:
> What does irony need the build directory for? I've only just looked at this tiny piece of the package.
>  Hide all checks

To find the compilation database. This could be the common denominator for the compilation databases that could be put in `irony-cdb` directly instead of on a per-compilation database basis. libclang Compilation Database API takes a buildDir as argument, not the path to the JSON Compilation Database file.
Additionally this could be a good default directory for `M-x compile`.

MikeSanMichel2:
I've pushed the search directories into a defcustom alist and added a wrapper around `locate-dominating-file` that makes use of it. I've also kept the "auto-reload" one liner you've already
signed of on.

Let's leave the changes to add-compile-path to another PR. Merging this would already be
a big improvement.

irony-cdb-json.el
MikeSanMichel2: I lied. I'm :-1::-1:  on breaking someone's working setup. But it's your show, so I split this off into a separate PR. It's an independent change.

irony-cdb-clang-complete.el
```diff
@@ -40,8 +40,12 @@
 
 (defun irony-cdb-clang-complete--locate-db ()
   (when buffer-file-name
-    (irony--awhen (locate-dominating-file buffer-file-name ".clang_complete")
-      (concat (file-name-as-directory it) ".clang_complete"))))
+    (irony--awhen
```
Sarcasm: I'm not sure I agree on adding this feature to `.clang_complete`. Unless proven otherwise, a `.clang_complete` should be found in a parent directory of the source file. Let's keep it that way for now.

irony-cdb.el
```diff
@@ -63,6 +63,20 @@ for files that it cannot handle."
   :type '(repeat function)
   :group 'irony-cdb)
 
+(defcustom irony-cdb-search-directory-list '("." "build")
```
Sarcasm: Why "." is needed?

I would remove "build" from the default, I'm not convinced about its general adoption. Also I prefer these directories to be opt-in, removing a directory is often difficult.

irony-cdb.el
```diff
@@ -114,6 +128,24 @@ for files that it cannot handle."
...
+
+Looks up the directory hierarchy from FILE for to locate any directory
+in `subdirectories` which contains NAME. Stops at the first parent
+directory where the test succeeds and returns the full path to file
+found."
+  (let ((check-subdirs
+         (lambda (dir)
+           (when dir
+             (cl-loop for subdir in subdirectories
+                      for match =(concat dir "/" subdir "/" name)
```
Sarcasm: formatting: I would add a space between `=` and `(`.
use `(file-name-as-directory FILE)` instead of hardcoding slashes.

irony-cdb.el
```diff
@@ -114,6 +128,24 @@ for files that it cannot handle."
...
 ;;
 
+(defun irony-cdb--locate-dominating-file-with-dirs (file name subdirectories)
+  "Convenience wrapper around `locate-dominating-file'
+
+Looks up the directory hierarchy from FILE for to locate any directory
+in `subdirectories` which contains NAME. Stops at the first parent
+directory where the test succeeds and returns the full path to file
+found."
+  (let ((check-subdirs
```
Sarcasm: I don't understand the need of this `check-subdirs` variable, can't you just loop over the directories directly?

irony-cdb.el
MikeSanMichel2: "." is the entry which covers the current behavior. Instead of special-casing it, check-subdirs loops over all defined directories. 

Removing build: I understood you agreed "build" was common enough. I think it is.
I disagree with you that customizing options is difficult. I also disagree that having extra
directories by default is something to be avoided .  Are you concerned about a scenario where:
1) a directory named `build` exists
2) it contains a json cdb.
3) **and** the user actually wants to use another cdb, which isn't
located in either the same directory or a parent directory.

....? and you believe that situation affects more users then those who use CMake and follow the `build/` convention which will now have irony working out of the box?

irony-cdb.el
MikeSanMichel2: 1. No. the behavior is to find the matching subdirectory in the nearest parent that matches.
if I `(locate-dominating dir ".")` and locate `../../<>.json` finds one 2 levels up, It won't check for the `one ../build/<>.json`. Doing it this way I find the closest one, and potentially check less directories.

2. `locate-dominating-file` returns the directory, not the path. So I have to reuse this check 
to figure out in which subdir the file was found. Since I reuse the logic (and it's not a one liner), this is simply DRY.

irony-cdb-clang-complete.el
MikeSanMichel2: Done.

irony-cdb.el
MikeSanMichel2: Done.

MikeSanMichel2:
Can you please disable AppVeyor until you get a chance to fix up the windows CI?

Sarcasm:
> Can you please disable AppVeyor until you get a chance to fix up the windows CI?

Second tentative, to ignore it. If it happens again I will use a more radical approach. Sorry about that.

irony-cdb-json.el
```diff
@@ -48,8 +48,31 @@ source directory.")
     (get-compile-options (irony-cdb-json--get-compile-options))))
 
 ;;;###autoload
-(defun irony-cdb-json-add-compile-commands-path (project-root
-                                                 compile-commands-path)
+(defun irony-cdb-json--find-common-root-directory (compile-commands-path)
```
Sarcasm: This part was moved to #239 I believe, I'm still undecided about it, better to leave it there.

irony-cdb.el
```diff
@@ -60,9 +54,29 @@ list of cons where the first element is a set of compile options
 and the second element the working directory expected for these
 commands. The compilation database should return an empty list
 for files that it cannot handle."
+  :group 'irony)
```
Sarcasm: I think the group is a great addition but I'm not sure about moving the documentation.

company-mode has its documentation in `company-backends` I tried to mimic that here.

Right now it's possible to get this documentation either by `describe-variable irony-cdb-compilation-databases` or via  the customize interface. To my knowledge, with this change it is only accessible via `customize-group irony-cdb`.

irony-cdb-json.el
MikeSanMichel2: Sorry, me and and magit are having a big fight this week.

irony-cdb.el
MikeSanMichel2: This was #239 too, but I'll change it back there.

MikeSanMichel2:
votre tentative est un échec...

Sarcasm:
Should be fixed now, if it still happens I will take a plane and talk to them!

irony-cdb.el
Sarcasm: > 3) and the user actually wants to use another cdb, which isn't located in either the same directory or a parent directory.
> 
> ....? and you believe that situation affects more users then those who use CMake and follow the build/ convention for which now irony will work out of the box?

I don't believe this, I believe it will happen mostly to me.


> I would actually favor the opposite approach. Let's add a check to verify (buffer-file-name) is included in the cdb before we accept it and then be very permissive about where we look by default. If there are any semi-popular conventions on where the build directory lives, just throw'em all in there.

Let's keep it simple and NOT do that for now. If someone encounter the issue we will discuss it then.

irony-cdb.el
Sarcasm: Ok, I think I'm super-confused by the 2 appearances of `check-subdirs`.

IIUC you are now using the version of `locate-dominating-file` where name is a function and not a path. I would prefer to avoid using this one since it is only available in Emacs 24. The last Ubuntu LTS is still using Emacs 23.4, we should support it.

irony-cdb.el
MikeSanMichel2: So, do you insist on removing `build`? I'll do it if you are.

irony-cdb.el
MikeSanMichel2: Good catch, I did not know about version availability. hmm.

Ok. We iterate over `irony-cdb-search-directory-list`, call `locate-dominating-file` for each 
one and end up with a list of 0 or more candidates.

How would you like to choose which one? in most cases, there'll only be one. If there are more,
the one with the most delimiters is a good heuristic for "most specific", or we could just take the first one and punt. Which do you prefer?

irony-cdb.el
Sarcasm: Sorry about that, I'm okay to keep build, and `.`, you made me change my mind.

MikeSanMichel2:
kermit_the_frog

irony-cdb.el
Sarcasm: > How do we choose which one to use? First, in most cases there'll be at most one so this isn't critical. If there are more then 1 we can choose the one with the most delimiters as a good heuristic for "most specific". Or we could just take the first one and punt on the whole thing.

Not sure to understand "the most delimiters".

> Which do you prefer?

I think you got the logic right in a previous comment:

> > 1. No. the behavior is to find the matching subdirectory in the nearest parent that matches.
if I (locate-dominating dir ".") and locate ../../<>.json 2 levels up, It won't find the nearer ../build/<>.json. Doing it this way finds the closest one and potentially performs less checks overall.

We have to find the closest one. Sub-dir when composed should be preferred to path closest to the root in the hierarchy.

I mean if:

```el
 (setq irony-cdb-search-directory-list ("." "build" "out/x86_64"))
```
Assuming `default-directory` / `buffer-file-name directory` is `~/foo/bar/qux.cpp`:

1. `~/foo/bar/out/x86_64/compile_commands.json`
2. `~/foo/build/compile_commands.json`
3. `~/compile_commands.json`

MikeSanMichel2: The most delimiters - the deepest directory.

irony-cdb.el
MikeSanMichel2: Thanks.

irony-cdb.el
MikeSanMichel2: Ok, I'm not sure I understood you perfectly, but see what you think.

irony-cdb-clang-complete.el
Sarcasm: Did you undone it?

MikeSanMichel2: > Did you undo it?

Yep.

Sarcasm:
I see that you have a WIP: tests.

What do you think of adding a test for `irony-cdb--locate-dominating-file-with-dirs` to this file: https://github.com/Sarcasm/irony-mode/blob/master/server/test/elisp/irony-cdb-json.el

Then I can try to give you some test with the result I expect.

MikeSanMichel2:
That's there only to help you understand what irony-cdb--choose-closest-path is doing

MikeSanMichel2:
How do I run the test suite?

Sarcasm:
`make check`, that's what travis do after the build is successful: https://github.com/Sarcasm/irony-mode/blob/3d7fe7cd43a070a57527d7a8fd1b98bc086848b8/.travis.yml#L32

MikeSanMichel2:
Ok, created some tests for `irony-cdb--choose-closest-path`. Plase add the tests for the behavior of locate. Try to match what it already does, if at all possible...

Sarcasm:
Here is a test that I would like to pass but fail currently:

```el
;; ugly but I think the current API of `irony-cdb--choose-closest-path' permit
;; to test this, it looks like to me the function should be aware of the
;; sub-dirs to make the right choice
(ert-deftest cdb/locate-dominating-file-with-dirs/foo ()
  (make-directory "/tmp/foo/bar/" t)
  (write-region "" nil "/tmp/foo/bar/qux.cpp")
  (make-directory "/tmp/foo/bar/out/x86_64" t)
  (write-region "" nil "/tmp/foo/bar/out/x86_64/cdb")
  (write-region "" nil "/tmp/foo/cdb")
  (should
   (string= "/tmp/foo/bar/out/x86_64/cdb"
          (irony-cdb--locate-dominating-file-with-dirs "/tmp/foo/bar/qux.cpp"
                                                       "cdb"
                                                       '("." "out/x86_64")))))
```

Do you agree/understand?

MikeSanMichel2:
There's a better way to test. So, you want the deepest directory after all?

Sarcasm:
Not really the deepest I think.
It's just, if I find a member of `irony-cdb-search-directory-list` in the current directory it should have higher priority than a member in `..`.

MikeSanMichel2:
How about now?

irony-cdb.el
```diff
@@ -114,6 +128,57 @@ for files that it cannot handle."
...
+  (when (listp paths)
+    (let ((paths (or
+                  ;; if we find a cdb in cwd or below, don't consider other candidates
+                  (cl-remove-if-not (lambda (x) (string-prefix-p (file-name-directory file) x)) paths)
+                  paths)))
+      (cl-loop for path in paths
+         with best-depth-delta = 999999 ; start at +inf
+         with best-path = nil ; we keep the best so far here
+         ;; all candidates have their depth compared to that of target file
+         with file-depth = (length (split-string file "/")) ;
```
Sarcasm: hardcoded unix-style path separator

irony-cdb.el
```diff
@@ -114,6 +128,57 @@ for files that it cannot handle."
...
+For remaining candidates, \"nearest\" is measured as abs. difference
+in path depth.
+- We prefer deeper paths at level +N to those at level -N.
+- If multiple paths are equally good, we return the last one.
+
+Returns nil if paths isn't a list of at least one element.
+"
+  (when (listp paths)
+    (let ((paths (or
+                  ;; if we find a cdb in cwd or below, don't consider other candidates
```
Sarcasm: or below? care to elaborate, to me it looks like we will favor a cdb in the current directory and that's all.
even if `irony-cdb-search-directory-list` does not specify it as a first element. I think we should not special case the current dir and just take the first best match from `irony-cdb-search-directory-list`, which defaults to `.` for the best match.

irony-cdb.el
MikeSanMichel2: Nope, that's an emacs-style path-seperator:

https://www.gnu.org/software/emacs/manual/html_node/elisp/File-Names.html

     On MS-DOS and MS-Windows, these functions (like the function that actually operate 
     on files) accept MS-DOS or MS-Windows file-name syntax, where backslashes separate 
     the components, as well as Unix syntax; **but they always return Unix syntax**. This 
     enables Lisp programs to specify file names in Unix syntax and work properly on all 
     systems without change.

`file` here is the result of `(buffer-file-name)` and paths come from `locate-dominating-file`.

irony-cdb.el
MikeSanMichel2: > "or below"? care to explain? to me it looks like we will favor a cdb in the current 
> directory and that's all.

Nope. The test is `(lambda (x) (string-prefix-p (file-name-directory file) x))` and any path
which lives in or under that directory will match, by definition.

> even if irony-cdb-search-directory-list does not specify it as a first element. I think we should > not special case the current dir and just take the first best match from 
> irony-cdb-search-directory-list, which defaults to . for the best match.

The current code already does exactly what you asked me for and passes the test you provided (`children-first` is an adapted version of it). 

Please provide a failing test if you've found a bug.

MikeSanMichel2: I just reread your comment for the 3rd time to make sure,  and I think I've finally been able to repunctuate it it makes sense to me.

> "or below"? care to explain? 

Sure.  The test is `(lambda (x) (string-prefix-p (file-name-directory file) x))` and any path
which lives in or under that directory will match, by definition. I believe the `children-first` test (based on the test you provided)  shows this works.

> to me it looks like we will favor a cdb in the current directory 
> even if irony-cdb-search-directory-list does not specify it as a first element. 

True, that's what I understood you wanted when you said earlier:
> It's just, if I find a member of irony-cdb-search-directory-list **in the current directory** it 
> should have higher priority than a member in .., no matter how deep the member is.
 
> I think we should not special case the current dir and just take the first best match from 
> irony-cdb-search-directory-list, which defaults to . for the best match.

That's exactly what the *first version* I wrote did and I threw that one away because we wanted:  
`/foo/bar/build/cdb`  (found by "build/", **one** level up)
to win over 
`/cdb` (found by ".", **two** levels up)

...when the search begins at `/foo/bar/x.cpp`. See https://github.com/Sarcasm/irony-mode/pull/237#discussion_r41975148.
 
Here's what it was, despite the problem I just mentioned is this what you want?

```
(defun irony-cdb--locate-dominating-file-with-dirs (filename start-path dirs)
  "Generalizes `locate-dominating-file` to search in given subdirectory list"
  (cl-loop for dir in dirs
           for match = (locate-dominating-file
                        start-path
                        (concat dir "/" filename))
           when match
           return (expand-file-name
                   (concat (mapconcat 'file-name-as-directory (list match dir)  "")
                           filename))
           ))

```

irony-cdb.el
Sarcasm: Wow, very nice, I had the feeling something like this was happening but apart from discovering that the `directory-sep-char` variable had been removed I didn't know what existed to replace it.

Thanks, sorry for all the trouble. :)

irony-cdb.el
Sarcasm: > which lives in or under that directory will match, by definition. I believe the children-first test (based on the test you provided) shows this works.

Yes you are right, I just don't get it for some reasons.



> That's exactly what the first version I wrote did and I threw that one away because (I thought) we wanted:
> 
> /foo/bar/build/cdb (found by "build/", one level up)
> to win over 
> /cdb (found by ".", two levels up)
> 
> ...when the search begins at /foo/bar/x.cpp. See #237 (comment).
> 
> Git reflog still had it - is this what you want?

No I like the current behavior, it is what I want. I just don't get all of the logic but it's not big deal.

MikeSanMichel2: You're maintaining it, you have to understand it.
Explain again what isn't clear, maybe I need to write clearer code.

Sarcasm:
I merged manually your commits. :)

Thank you, sorry for being a pain during this review, I really appreciate what you did!

irony-cdb.el
MikeSanMichel2: `buffer-file-name` isn't strictly one of the functions in that section, so I added 2c53d40 to be safe.

MikeSanMichel2:
Hallelujah!

No worries, review can be annoying but you always end up with something better. That's how it works.
If you just missed my last comment , you should cherry-pick 2c53d40.

Sarcasm:
Done, thank you.

irony-cdb.el
MikeSanMichel2: I just saw this, guess we made the right choice:

$ cat cmake/utils.cmake 
 
> function(check_for_in_source_build)
>   if (CMAKE_SOURCE_DIR STREQUAL CMAKE_BINARY_DIR AND NOT MSVC_IDE)
>     message(FATAL_ERROR "In-source builds are not allowed.
>
>   **Please create a build/ directory** and run cmake from there, passing
>   the path to this source directory as the last argument. This process
>   created the file `CMakeCache.txt' and the directory `CMakeFiles'.
>   Please delete them.
> ")
>   endif()
> endfunction()

Looks like it really is the "official" place to build things.

Sarcasm: This is from me, copy-pasted from LLVM cmake build, I would not go as far as to say one project makes it "official". But yes this is a sound guess. :)

For example, on QtCreator, the build is often like this by default I believe:

* foo/project/CMakeLists.txt
* foo/project-qtcreator-build

Something along these lines.

MikeSanMichel2: oh,I thought it was from cmake upstream,  but LLVM matters as well. Also, at least irony
supports its own recommendations now... :)

PR #239, opened 2015-10-11
irony-cdb-json-add-compile-commands-path guesses project root from cdb

MikeSanMichel2:
Changes argument order -- breaking someone's .emacs.

Continuing #237


MikeSanMichel2:
I'm still -1 on this breaking change and although guessting the project root from the cdb contents is nice,  now that #237 is merged I would just adjust the search path instead.

Sarcasm:
Fine with me.

PR #257, opened 2015-12-06
ignore errors in server response processing

jerryxgh:
In function `irony--server-process-filter`, when multiple responses are
returned, will call `irony--process-server-response` to process theme on
by one. But if some one of the responses (not the last) signals error in
processing, then `irony--server-process-filter` will be interrupted,
that will leave some responses not processed, and corresponding
callbacks in the process not poped. Then follow-up requests will call
wrong callbacks and waiting in `accept-process-output` forever.

Run `irony-server-kill` manually will clear the callbacks and fix this,
but ignore errors in server response processing will be better

### DIFF
--- a/irony.el
+++ b/irony.el
@@ -588,7 +588,8 @@ list (and undo information is not kept).")
 (defun irony--process-server-response (process response)
   (let ((sexp (read response))
         (callback (irony--server-process-pop-callback process)))
-    (apply (car callback) sexp (cdr callback))))
+    (ignore-errors
+      (apply (car callback) sexp (cdr callback)))))
 
 (defun irony--server-process-filter (process output)
   "Handle output that come from an irony-server process."
--- a/irony.el
+++ b/irony.el
@@ -588,7 +588,7 @@ list (and undo information is not kept).")
 (defun irony--process-server-response (process response)
   (let ((sexp (read response))
         (callback (irony--server-process-pop-callback process)))
-    (ignore-errors
+    (with-demoted-errors "Warning: %S"
       (apply (car callback) sexp (cdr callback)))))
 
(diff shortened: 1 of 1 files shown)


Sarcasm:
thanks, do you think you could print the error or still allow the error to be debugged when toggle-debug-on-error is active? There are functions for this such as `with-demoted-errors` or `condition-case-unless-debug` I think that could make it easier to debug these errors.

jerryxgh:
`ignore-errors` will not print the error or allow the error to be debugged when toggle-debug-on-error is active, so you are right, `with-demoted-errors` is appropriate here.

jerryxgh:
Would you please merge this pull request? When I update this package, I encounter this bug again(by diff file in magit), and emacs become almost unavailable. Read the code I found this request is not merged.

So thank you for the trouble.

Sarcasm:
It's merged, sorry for the delay.

By the way, errors should not appear here I believe, do you know what was is the error your get?

jerryxgh:
This is the error debug stack, happens when using diff on a c source file in magit:

Debugger entered--Lisp error: (wrong-type-argument stringp nil)
  expand-file-name(nil)
  flycheck-same-files-p("path-to-a-c-source-file" nil)
  flycheck-relevant-error-p()
  flycheck-relevant-errors()
  flycheck-finish-current-syntax-check()
  flycheck-report-buffer-checker-status()
  irony-diagnostics--request-handler()
  irony--process-server-response()
  irony--server-process-filter()

Sarcasm:
ok, thanks, it may be a flycheck issue or maybe irony is not doing something right.

jerryxgh:
In my opinion, it should be a flycheck issue that woking with magit.

PR #261, opened 2015-12-15
substitute rpath with full libclang path on OS X

kevinushey:
The goal here is to ensure that e.g. `@rpath/libclang.dylib` is substituted with the actual `libclang` path discovered at install time, as otherwise `irony-server` will fail to start (unless the rpath is actually set to a path with the libclang library, which I think if often not the case by default).

Note that I tried fiddling with all the rpath-related variables described at https://cmake.org/Wiki/CMake_RPATH_handling, but had no joy in otherwise forcing CMake to generate the binary with an absolute library path.

Let me know what you think -- it also might be better to hide this configuration behind an option, but I think this will help ensure that `irony-server` 'just works' when installed e.g. through `(irony-install-server)` from Emacs.

### DIFF
--- a/server/cmake/modules/FindLibClang.cmake
+++ b/server/cmake/modules/FindLibClang.cmake
@@ -61,20 +61,20 @@ endforeach()
 
 find_path(LIBCLANG_INCLUDE_DIR clang-c/Index.h
   PATHS ${libclang_llvm_header_search_paths}
-  PATH_SUFFIXES LLVM/include #Windows package from http://llvm.org/releases/
+  PATH_SUFFIXES llvm/include #Windows package from http://llvm.org/releases/
   DOC "The path to the directory that contains clang-c/Index.h")
 
 # On Windows with MSVC, the import library uses the ".imp" file extension
 # instead of the comon ".lib"
 if (MSVC)
   find_file(LIBCLANG_LIBRARY libclang.imp
-    PATH_SUFFIXES LLVM/lib
+    PATH_SUFFIXES llvm/lib
     DOC "The file that corresponds to the libclang library.")
 endif()
 
 find_library(LIBCLANG_LIBRARY NAMES libclang.imp libclang clang
   PATHS ${libclang_llvm_lib_search_paths}
-  PATH_SUFFIXES LLVM/lib #Windows package from http://llvm.org/releases/
+  PATH_SUFFIXES llvm/lib #Windows package from http://llvm.org/releases/
   DOC "The file that corresponds to the libclang library.")
 
 get_filename_component(LIBCLANG_LIBRARY_DIR ${LIBCLANG_LIBRARY} PATH)
--- a/server/src/CMakeLists.txt
+++ b/server/src/CMakeLists.txt
@@ -68,3 +68,12 @@ set_source_files_properties(main.cpp
 target_link_libraries(irony-server ${LIBCLANG_LIBRARIES})
 
 install(TARGETS irony-server DESTINATION bin)
+
+# don't use @rpath
+if (APPLE)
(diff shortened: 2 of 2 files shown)


server/cmake/modules/FindLibClang.cmake
```diff
@@ -61,20 +61,20 @@ endforeach()
 
 find_path(LIBCLANG_INCLUDE_DIR clang-c/Index.h
   PATHS ${libclang_llvm_header_search_paths}
-  PATH_SUFFIXES LLVM/include #Windows package from http://llvm.org/releases/
+  PATH_SUFFIXES llvm/include #Windows package from http://llvm.org/releases/
```
Sarcasm: did the windows package changed? on my computer it was LLVM uppercase, I copy-pasted it.

kevinushey: It looks like it is indeed `LLVM` on Windows but `llvm` on OS X / other platforms -- this is based on the binaries available at http://llvm.org/releases/download.html?

kevinushey: Sorry, I should add some motivation for the change as well -- on my OS X machine, it was indeed discovering a `libclang.dylib` library at `/usr/local/llvm/lib/libclang.dylib`, but persisted the capital `LLVM` despite that the directory was indeed `llvm`. This might affect users with case-sensitive filesystems (although CMake might just automatically normalize the path in that case, I am not sure)

Sarcasm: irony-server was unable to build without this change or you just saw that it was lowercase on your system and changed?

We may need both variant eventually.

kevinushey: Irony was still able to build (on OS X) with upper-case `LLVM`, although on my OS X system the path was indeed `llvm`.

Sarcasm: Sorry but would you mind telling me what package you installed? I downloaded 2 packages from http://llvm.org/releases/download.html:

1. clang+llvm-3.7.0-amd64-unknown-freebsd10
2. clang+llvm-3.7.0-x86_64-apple-darwin

None of them provide a path such as `llvm/include`.

If you installed one of this package a decided to put it in `/usr/local/llvm` that's not standard, not more standard than putting it in `/opt/my-llvm-install/`.

In this case, you have to tell CMake about your non-standard path with `-DCMAKE_PREFIX_PATH=/opt/my-llvm-install`.

On Windows, the packages installs in `C:/Program File/LLVM/` I think, and I used the PATH_SUFFIXES for that, even though I'm am not sure it is the proper way to do but the PATH_SUFFIXES.

server/src/CMakeLists.txt
```diff
@@ -68,3 +68,12 @@ set_source_files_properties(main.cpp
 target_link_libraries(irony-server ${LIBCLANG_LIBRARIES})
 
 install(TARGETS irony-server DESTINATION bin)
+
+# don't use @rpath
+if (APPLE)
+  add_custom_command(
+    TARGET irony-server
+    POST_BUILD COMMAND
+    ${CMAKE_INSTALL_NAME_TOOL} "-change" "@rpath/libclang.dylib" "${LIBCLANG_LIBRARY}" "$<TARGET_FILE:irony-server>"
```
Sarcasm: which package are you using? I don't think it is needed for some other packages.

Wouldn't be better to put this in the wiki or at the end of the README in the FAQ/troubleshooting section?

kevinushey: I think you're right. I'll close this PR and instead make one adding this information to the README. Sorry for the various other noise.

PR #263, opened 2015-12-19
instructions on using install_name_tool

kevinushey:
This provides some instructions on using `install_name_tool`, as another means of resolving issues where `irony-server` might not be able to discover the `libclang.dylib` library.


Sarcasm:
thanks

PR #273, opened 2016-01-15
Fix #272 - no longer fails when ccache is used

### LINKED ISSUE #272: Fails when the compiler is prefixed with `ccache`
One way to use ccache with cmake is to list it before the compiler name when calling CMake:

    CC='ccache clang' CXX='ccache clang++' cmake ...

That means that the commands in the compilation database start with `ccache clang++`. When reading it Irony correcly takes out the compiler (the first parameter) in order to get the options. But if ccache is preceding it, it ends up passing the compiler itself as part of `compile_options` in `irony--send-parse-request`. As a result, the compilation fails and no autocompletion happens.

atilaneves:


atilaneves:
Ping @Sarcasm

Sarcasm:
merged in manually. Thanks for the ping ;)

atilaneves:
No problem :)

Issue #274, opened 2016-01-25
irony-mode doesn't work


windrg:
Hi,

First of all, I really appreciate your greate work.
However I couldn't activate it properly.
It seems to work in a small project but it doesn't work at all in a large project like linux kernel.
Could you please take a look at my config as below?

```
	(add-hook 'after-init-hook 'global-company-mode)

	(custom-set-variables '(company-clang-executable "/usr/bin/clang-3.5"))


	(add-hook 'c++-mode-hook 'irony-mode)
	(add-hook 'c-mode-hook 'irony-mode)
	(add-hook 'objc-mode-hook 'irony-mode) ;; do I need this now?



	(defun company-complete-common-or-previous-cycle()
	  "Insert the common part of all candidates, or select the previous one."
	  (interactive)
	  (when (company-manual-begin)
	    (let ((tick (buffer-chars-modified-tick)))
	      (call-interactively 'company-complete-common)
	      (when (eq tick (buffer-chars-modified-tick))
		(let ((company-selection-wrap-around t))
		  (call-interactively 'company-select-previous))))))

	(eval-after-load 'company
	  '(progn 
	    ;; (define-key company-active-map (kbd "TAB") 'company-select-next)
	    ;; (define-key company-active-map [tab] 'company-select-next)
	    (define-key company-active-map (kbd "<tab>") 'company-complete-common-or-cycle)
	    (define-key company-active-map [tab] 'company-complete-common-or-cycle)
	    (define-key company-active-map (kbd "<backtab>") 'company-complete-common-or-previous-cycle)
	    (define-key company-active-map [backtab] 'company-complete-common-or-previous-cycle)
	    ;; FIXME how to bind(or override ) ESC w/ company-abort???
	    ;; now it is bound to evil-normal-state or keyboard-quit.
	    ;; (define-key company-active-map (kbd "<esc>") 'company-abort)
	    ;; (define-key company-active-map ["\e"] 'company-abort)

	    (setq company-tooltip-limit 500)                      ; bigger popup window
	    (setq company-idle-delay .3)                         ; decrease delay before autocompletion popup shows
	    (setq company-echo-delay 0)                          ; remove annoying blinking
	    (setq company-begin-commands '(self-insert-command)) ; start autocompletion only after typing
	    ) )
	    ;; TODO try this company w/ Helm?? maybe it could be a faster way to input

	;; replace the 'completion-at-point' and 'complete-symbol' bindings in
[...]

PR #281, opened 2016-02-11
Add support for shell commands in .clang_complete

alexmurray:
Any form in .clang_complete contained in backticks (``) will
be evaluated as a shell command. This allows forms such as:

```
-I/lib/modules/`uname -r`
```

to refer to the currently installed Linux kernel module directory.

See Issue #274 for motivation.

### DIFF
--- a/irony-cdb-clang-complete.el
+++ b/irony-cdb-clang-complete.el
@@ -46,6 +46,14 @@
 (defun irony-cdb-clang-complete--load-db (cc-file)
   (with-temp-buffer
     (insert-file-contents cc-file)
+    ;; shell eval any forms in backticks
+    (save-excursion
+      (while (re-search-forward "`\\(.*?\\)`" nil t)
+        (replace-match
+         ;; strip newlines
+         (replace-regexp-in-string "\n\\'" ""
+                                   (shell-command-to-string
+                                    (match-string 1))))))
     (list
      (cons
       ;; compile options with trailing whitespaces removed


Sarcasm:
Hi,

I'm not too keen on modifying the `.clang_complete` format with something like this, I see the future as "hey it's not safe, someone can put malicious commands in here, I wrote some code to guard against this...". Also this is used by some other projects, so I believe it's best to keep the way it is.

Also I think it's better to have compilation-database tools the config.

IMHO, right now a better entry point for such things is to write a custom cdb.

alexmurray:
Ok fair enough - I agree it does create a vector for remote code execution - and I also don't like the idea of a whitelist etc as maintenence of the would be too cumbersome. However, even writing a custom cdb will result in the same problem - ie. still having a vector for RCE - so perhaps it is not worth pursuing?

Sarcasm:
Yeah, I think it can be written locally first and if the person that uses this feel like it is useful and want to share it with irony I'm all for it. But I have the feeling we can do without if better compilation database toolings comes out.

alexmurray:
So I came up with an external solution just using the shell and a `Makefile` to generate a `.clang_complete` from a template (so that can be used instead of this PR):

For example the following can be used as a generic template for a `.clang_complete` for compiling a Linux kernel module:

`.clang_complete.in`:
```
-nostdinc
-I/lib/modules/`uname -r`/build/include
-I/lib/modules/`uname -r`/build/include/uapi
-I/lib/modules/`uname -r`/build/arch/x86/include
-I/lib/modules/`uname -r`/build/arch/x86/include/generated
-I/lib/modules/`uname -r`/build/arch/x86/include/uapi
-I/lib/modules/`uname -r`/build/arch/x86/include/generated/uapi
-I.
-D__KERNEL__
-DCONFIG_AS_CFI=1
-DCONFIG_AS_CFI_SIGNAL_FRAME=1
-DCONFIG_AS_CFI_SECTIONS=1
-DCONFIG_AS_FXSAVEQ=1
-DCONFIG_AS_AVX=1
-DCONFIG_AS_AVX2=1
-DKBUILD_MODNAME=foo
-DMODULE
-D"KBUILD_STR(s)=#s"
-include /lib/modules/`uname -r`/build/include/linux/kconfig.h
```

And with a simple Makefile rule we can generate the actual `.clang_complete` with the backticks evaluated:

`Makefile`:
```
.clang_complete: .clang_complete.in
	        while read line; do line=$$(echo echo $$line | sh); echo $$line >> .clang_complete; done < $^
```

PR #293, opened 2016-03-12
Guess compile commands in irony-server.

Hylen:
Using boost property_tree to read the database, and cache the database
between subsequent reads. For guessing commands, path operations from
boost filesystem are used. If boost isn't detected on the system, the
database is disabled.


Hylen:
I've removed rapidjson in favour of property_tree, but I haven't removed the test that you didn't like before. If you still think I should remove it, please give me a call and I'll update the PR.

PR #294, opened 2016-03-15
Look at table first when searching for database.

Hylen:
Previously the dominating directories were searched first when looking
for a compile_commands.json file. This could be confusing to a user,
having specified one database and still be using another one.


Sarcasm:
Thank you!

Issue #302, opened 2016-05-05
Exposing the cdb project alist


netromdk:
Hello. First off, irony-mode is the first mode to give me correct completions for my very big work projects. So thanks for that!

Secondly, the mode works great when you only have one build folder for your project but at work I often have several build folders for different branches (to not having to recompile too much too often) with varying includes etc. Since the project root is the same for those build folders then it will always use the first value in `irony-cdb-json--project-alist`. So to switch between I need to put the one I want at the top of the list (or just before any of the others with the same project root).

Now, `irony-cdb-json--project-alist` is private so I can't just access it without autoloading it first but I was wondering if you might consider exposing it in your package? I would like to write a package that lists the known cdbs and then choosing the one to make active.

Preferably also `irony-cdb-json--save-project-alist` so the values would be saved back to the file in the same way, if the order has been changed.

Made a working demo when those two symbols are exposed: https://github.com/netromdk/irony-selector
I think it could be interesting to add to melpa.

netromdk:
Preferably also `irony-cdb-json--save-project-alist` so the values would be saved back to the file in the same way, if the order has been changed.

Sarcasm:
I'm not totally against exposing this variables but to be frank, if I find energy to rework on the compilation database stuff these variables might totally disappear, evolve, ...So making them public may be a lie. If there was no external dependency (helm) we could integrate `irony-selector` in irony, and that would means I could find way to keep this working even if reworking the cdb.

I'm wondering if instead of choosing the first one, we could chose the most recent one instead?
Selecting a new compilation database on a CMake project would be semi-automatic, one a build command has been run the compile_commands.json should have been updated. This may not be perfect but if that solves 90% of the needs that may be worth it.
---
On a side note, this make me think about something related.
The design of the compilation database is that you can have several compile commands for one file. Right now the first one is proposed.

netromdk:
I understand. I could make a PR to put it into your `irony-cdb-json` package and then exposing handles like in lieu of how yasnippet has `yas-prompt-functions` where you supply a function like `(defun thepromptfunc (prompt choices &optional display-fn)..)`? That way I could provide a generic, non-helm default but making it possible to add it myself anyway. Or include the helm variant that only is used if helm is loaded.

Choosing the most recent is interesting but sometimes I switch between branches/build folders without actually running cmake or invoking build commands until after having done some initial edits.

I didn't know you could have more in one file actually.

Sarcasm:
[...]

PR #303, opened 2016-05-06
Implemented irony-cdb-json-select.

netromdk:
Select CDB to use with a prompt, which is useful when having several CDBs for the same project root.

As we talked about in issue #302.


Sarcasm:
Merged manually, thank you!

PR #304, opened 2016-05-07
Implemented irony-cdb-json-select-most-recent.

netromdk:
It sorts the CDB list so that most recently modified files appear first, which ensures that it works even when having multiple projects with multiple build folders.

As discussed in issue #302.

### DIFF
--- a/irony-cdb-json.el
+++ b/irony-cdb-json.el
@@ -77,9 +77,6 @@ directories to project directory."
     (let ((elm (nth pos target-list)))
       (append (list elm) (delete elm target-list)))))
 
-(defun irony-cdb-json--cdb-list ()
-  (mapcar #'cdr irony-cdb-json--project-alist))
-
 (defun irony-cdb-json--choose-cdb ()
   "Prompt to select CDB from current project root."
   (let ((proot (irony-cdb-json--find-best-prefix-path
--- a/irony-cdb-json.el
+++ b/irony-cdb-json.el
@@ -77,12 +77,17 @@ directories to project directory."
     (let ((elm (nth pos target-list)))
       (append (list elm) (delete elm target-list)))))
 
+(defun irony-cdb-json--cdb-list ()
+  (mapcar (lambda (x) (cdr x)) irony-cdb-json--project-alist))
+
 (defun irony-cdb-json--choose-cdb ()
-  (let ((choices (mapcar (lambda (x) (cdr x)) irony-cdb-json--project-alist)))
-    (completing-read "Choose Irony CDB: " choices nil 'require-match nil)))
+  (let ((cdbs (irony-cdb-json--cdb-list)))
+    (if (> (length cdbs) 0)
+        (completing-read "Choose Irony CDB: " cdbs nil 'require-match nil)
+      nil)))
 
 ;;;###autoload
-(defun irony-cdb-json-select ()
+(defun irony-cdb-json-select (&optional cdb)
   "Select CDB to use with a prompt.
 
 It is useful when you have several CDBs with the same project
@@ -90,15 +95,42 @@ root.
 
 The completion function used internally is `completing-read' so
 it could easily be used with helm, for instance, by enabling
-`helm-mode' before calling the function."
+`helm-mode' before calling the function using an advice."
   (interactive)
-  (let ((pos (cl-position (irony-cdb-json--choose-cdb)
-                          irony-cdb-json--project-alist
-                          :test (lambda (x y) (string-equal x (cdr y))))))
-    (setq irony-cdb-json--project-alist
-          (irony-cdb-json--put-first pos irony-cdb-json--project-alist))
-    (irony-cdb-json--save-project-alist)
-    (irony-cdb-autosetup-compile-options)))
+  (let (the-cdb pos)
+    (setq the-cdb (or cdb (irony-cdb-json--choose-cdb)))
+    (if (or (null the-cdb)
+            (not (cl-member the-cdb (irony-cdb-json--cdb-list) :test 'string-equal)))
+        (message "CDB not in list: %s" the-cdb)
+      (progn
+        (setq pos (cl-position the-cdb
+                               irony-cdb-json--project-alist
+                               :test (lambda (x y) (string-equal x (cdr y)))))
+        (setq irony-cdb-json--project-alist
+              (irony-cdb-json--put-first pos irony-cdb-json--project-alist))
+        (irony-cdb-json--save-project-alist)
+        (irony-cdb-autosetup-compile-options)
+        (message "%s" (first irony-cdb-json--project-alist))))))
+
+(defun irony-cdb-json--last-mod (file)
+  (nth 5 (file-attributes file)))
+
+(defun irony-cdb-json--last-mod-cdb ()
+  (let (cdbs result)
+    (progn
+      (setq cdbs (irony-cdb-json--cdb-list))
+      (setq result (first cdbs))
+      (dolist (cdb cdbs)
+        (when (time-less-p (irony-cdb-json--last-mod result)
+                           (irony-cdb-json--last-mod cdb))
+          (setq result cdb)))
+      result)))
+
+;;;###autoload
+(defun irony-cdb-json-select-most-recent ()
+  "Select CDB that is most recently modified."
+  (interactive)
+  (irony-cdb-json-select (irony-cdb-json--last-mod-cdb)))
 
 (defun irony-cdb-json--get-compile-options ()
   (irony--awhen (irony-cdb-json--locate-db)
--- a/irony-cdb-json.el
+++ b/irony-cdb-json.el
@@ -78,38 +78,41 @@ directories to project directory."
       (append (list elm) (delete elm target-list)))))
 
 (defun irony-cdb-json--cdb-list ()
-  (mapcar (lambda (x) (cdr x)) irony-cdb-json--project-alist))
+  (mapcar #'cdr irony-cdb-json--project-alist))
 
 (defun irony-cdb-json--choose-cdb ()
-  (let ((cdbs (irony-cdb-json--cdb-list)))
-    (if (> (length cdbs) 0)
-        (completing-read "Choose Irony CDB: " cdbs nil 'require-match nil)
-      nil)))
+  "Prompt to select CDB from current project root."
+  (let ((proot (irony-cdb-json--find-best-prefix-path
+                (irony-cdb-json--target-path)
+                (mapcar 'car irony-cdb-json--project-alist)))
+        (cdbs '()))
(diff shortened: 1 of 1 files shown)


irony-cdb-json.el
```diff
@@ -77,28 +77,55 @@ directories to project directory."
     (let ((elm (nth pos target-list)))
       (append (list elm) (delete elm target-list)))))
 
+(defun irony-cdb-json--cdb-list ()
+  (mapcar (lambda (x) (cdr x)) irony-cdb-json--project-alist))
```
Sarcasm:
```el
(mapcar #'cdr irony-cdb-json--project-alist)
```

irony-cdb-json.el
```diff
@@ -77,28 +77,55 @@ directories to project directory."
...
+
 (defun irony-cdb-json--choose-cdb ()
-  (let ((choices (mapcar (lambda (x) (cdr x)) irony-cdb-json--project-alist)))
-    (completing-read "Choose Irony CDB: " choices nil 'require-match nil)))
+  (let ((cdbs (irony-cdb-json--cdb-list)))
+    (if (> (length cdbs) 0)
+        (completing-read "Choose Irony CDB: " cdbs nil 'require-match nil)
```
Sarcasm: Maybe only list the cdbs that are applicable to the current project? the cdb that matches the current-directory's project?

irony-cdb-json.el
netromdk: Thanks, I'll include it in next commit.

irony-cdb-json.el
netromdk: Good point. Do you already have a function to determine that? Otherwise I'll figure it out somehow.

irony-cdb-json.el
```diff
@@ -77,28 +77,55 @@ directories to project directory."
...
+(defun irony-cdb-json-select (&optional cdb)
   "Select CDB to use with a prompt.
 
 It is useful when you have several CDBs with the same project
 root.
 
 The completion function used internally is `completing-read' so
 it could easily be used with helm, for instance, by enabling
-`helm-mode' before calling the function."
+`helm-mode' before calling the function using an advice."
```
Sarcasm: I don't think that's the best way to do it.

For ido, I was able to temporary let-bind `completing-read-function` `ido-completing-read`, e.g:

```el
(let ((completing-read-function 'ido-completing-read))
    (irony-cdb-json-select))
```

How do you do for helm?

netromdk: I currently do it like this:
```elisp
(defadvice irony-cdb-json-select (around advice-irony-cdb-json-select activate)
    "Select CDB using helm prompt."
    (interactive)
    (let ((on (bound-and-true-p helm-mode)))
      (when (not on) (helm-mode 1))
      (unwind-protect
          ad-do-it
        (when (not on) (helm-mode -1)))))
```

irony-cdb-json.el
```diff
@@ -77,28 +77,55 @@ directories to project directory."
...
-  (let ((pos (cl-position (irony-cdb-json--choose-cdb)
-                          irony-cdb-json--project-alist
-                          :test (lambda (x y) (string-equal x (cdr y))))))
-    (setq irony-cdb-json--project-alist
-          (irony-cdb-json--put-first pos irony-cdb-json--project-alist))
-    (irony-cdb-json--save-project-alist)
-    (irony-cdb-autosetup-compile-options)))
+  (let (the-cdb pos)
+    (setq the-cdb (or cdb (irony-cdb-json--choose-cdb)))
+    (if (or (null the-cdb)
```
Sarcasm: Can we be silent when the-cdb is null?

irony-cdb-json.el
```diff
@@ -77,28 +77,55 @@ directories to project directory."
...
+            (not (cl-member the-cdb (irony-cdb-json--cdb-list) :test 'string-equal)))
+        (message "CDB not in list: %s" the-cdb)
```
Sarcasm: Calling `error` would be better than `message`.

```el
(error "Irony: invalid compilation database" the-cdb)
```

irony-cdb-json.el
```diff
@@ -77,28 +77,55 @@ directories to project directory."
...
+        (setq irony-cdb-json--project-alist
+              (irony-cdb-json--put-first pos irony-cdb-json--project-alist))
+        (irony-cdb-json--save-project-alist)
+        (irony-cdb-autosetup-compile-options)))))
+
+(defun irony-cdb-json--last-mod (file)
+  (nth 5 (file-attributes file)))
+
+;;;###autoload
+(defun irony-cdb-json-select-most-recent ()
```
Sarcasm: What happens the the file does not exists? Build directories have a habit to disappear and sometimes being recreated, I'm afraid the current code might complain.

irony-cdb-json.el
netromdk: Actually I think I'll remove the optional argument because it was only there because `irony-cdb-json-select-most-recent` used it in an intermediate commit. Then it doesn't need to have the error message at all. Agree?

irony-cdb-json.el
Sarcasm: `irony-cdb-json--find-best-prefix-path` might do something useful.

irony-cdb-json.el
netromdk: Just tested. No it doesn't behave well. Will improve that part.

irony-cdb-json.el
Sarcasm: Ok, it's because you want to enable then disable helm automatically. Then I guess it's fine to do this with an advice, but if people want an heml prompt it will work without the advice as long as the user has helm in his config? If so, remove the advice comment which might be confusing and is for more advanced stuff.

netromdk: It is because helm mode isn't turned on unless you're doing completions and therefore I only want to turn it on if it isn't already. But I'll remove that comment if it is confusing.

irony-cdb-json.el
Sarcasm: Agreed.

irony-cdb-json.el
Sarcasm: Hum, okay I will trust you on this.
Personnally I think making a new function is easier to understand (never used advices for now).

```el
(defun my-irony-cdb-json-select ()
  (interactive)
  (if (bound-and-true-p helm-mode)
     (irony-cdb-json-select)
  (helm-mode 1)
  (irony-cdb-json-select)
  (helm-mode -1)))
```

netromdk: Oh I did that myself first as well but if you do C-g to not choose anything then `helm-mode` stays on. That's why I did the `unwind-protect`. And used the advice so I don't have to call another function.

Sarcasm: Yes you are right about the unwind-protect.
For me an advice is to modify the code of/monkey patch a function.
So if helm provided `helm-completing-read`, I would expect to advice the function to replace `completing-read` by this function. Here I was just a little surprised.

netromdk:
I've made the improvements mentioned above. :)

irony-cdb-json.el
```diff
@@ -77,9 +77,22 @@ directories to project directory."
     (let ((elm (nth pos target-list)))
       (append (list elm) (delete elm target-list)))))
 
+(defun irony-cdb-json--cdb-list ()
```
Sarcasm: Seems like the function can be removed now.

netromdk: Removed it.

irony-cdb-json.el
```diff
@@ -77,9 +77,22 @@ directories to project directory."
...
+  (mapcar #'cdr irony-cdb-json--project-alist))
+
 (defun irony-cdb-json--choose-cdb ()
-  (let ((choices (mapcar (lambda (x) (cdr x)) irony-cdb-json--project-alist)))
-    (completing-read "Choose Irony CDB: " choices nil 'require-match nil)))
+  "Prompt to select CDB from current project root."
+  (let ((proot (irony-cdb-json--find-best-prefix-path
+                (irony-cdb-json--target-path)
+                (mapcar 'car irony-cdb-json--project-alist)))
+        (cdbs '()))
```
Sarcasm: I believe cdbs could be "more functionally set" by using a filter operation, e.g: cl-remove-if.

irony-cdb-json.el
```diff
@@ -77,9 +77,22 @@ directories to project directory."
...
+    (progn
+      (dolist (elm irony-cdb-json--project-alist)
+        (when (string-equal proot (car elm))
+          (setq cdbs (append cdbs (list (cdr elm))))))
+      (if (> (length cdbs) 0)
```
Sarcasm: don't compare the length, just do:

```el
(when cdbs
  ...)
```

irony-cdb-json.el
```diff
@@ -100,6 +114,30 @@ it could easily be used with helm, for instance, by enabling
...
     (irony-cdb-autosetup-compile-options)))
 
+(defun irony-cdb-json--last-mod (file)
+  (nth 5 (file-attributes file)))
+
+;;;###autoload
+(defun irony-cdb-json-select-most-recent ()
+  "Select CDB that is most recently modified."
+  (interactive)
+  (let (exists notexists)
```
Sarcasm: FYI `(let (exist notexists))` already initialize the member to `nil`, `nil` and `'()` are equivalent, so you don't need the 2 set lines.

And you would not want to do `let` + `setq`, just initialize in the `let`: `(let ((a '()) (b '()))`

irony-cdb-json.el
```diff
@@ -100,6 +111,30 @@ it could easily be used with helm, for instance, by enabling
     (irony-cdb-json--save-project-alist)
     (irony-cdb-autosetup-compile-options)))
 
+(defun irony-cdb-json--last-mod (file)
+  (nth 5 (file-attributes file)))
```
Sarcasm: You could simplify the handling of non-existing file by returning a null time value when the file does not exists:

```el
(or (nth 5 (file-attributes file))
      '(0 0 0 0))
```

irony-cdb-json.el
netromdk: Thanks :)

irony-cdb-json.el
netromdk: I actually tries this but got an error and got away from it again. I like this solution far better.

Sarcasm:
Hi,

Sorry did not take time to test yours changes but the code look simpler now, I will try to test that tomorrow. Thank you.

netromdk:
Sounds great, thanks :)

Sarcasm:
Merged as https://github.com/Sarcasm/irony-mode/commit/bbb3d4ac7d551d392619ca8a528c49bb29f1c98f

Thank you!

PR #311, opened 2016-05-31
irony-cdb-json: fix compiler flags when using ccache

arximboldi:
This commit changes the cleanup of the compiler flags to call `irony-cdb--remove-compiler-from-flags`. This function is used by the `libclang` database to properly remove the compiler from the flags. In particular: it takes into account if the compiler is wrapped by a call to `ccache`.

Similar changes might be needed in other databases to be able to properly support ccache.


arximboldi:
@Sarcasm Follow up to #272 

I hope this fix is appropriate, thanks for taking a look!

Sarcasm:
Cherry picked as https://github.com/Sarcasm/irony-mode/commit/6636e2ef4816d0cfa8771259a68f7385d056000c

Thanks.

PR #312, opened 2016-06-01
Improved diagnostics

itollefsen:
Adds spell checking, notes, hints and end column to Irony's diagnostic output.

This rewrites dumpDiagnostics() so that it also finds the end column,  includes notes and fixit hints.

The output format has not changed, but instead of outputting the offset, it now outputs the end column. This field was previously unused, but a new function as been added to irony-diagnostics.el to pick it up.

This rewrite has two goals; find the end column to speed up Flycheck's  error-region for flycheck-irony (separate project and change) and improve diagnostic reporting by including notes and fixit hints.

Notes are kept in child diagnostics, so these are now picked up and parsed as well.

Once parsed, notes are collapsed into their parent if on the same line (since it's mostly extra information about the parent).

Some more filtering is also done to get rid of reference to other files (which are irrelevant for flycheck-irony) and "declared here" notes that will appear out of context before the actual message (usually) when marked by Flycheck.

Messages are changed to also include notes (or'ed together, which seems to work quite well) and a hint, if there is only one and it's not attached to a note (which already spells out what the hint says).

Explanatory comments have been added in the code to clarify what it's trying to achieve.

-fspell-check has been added to the flags that TUManager uses to get libclang to perform spell
checking. This has a performance penalty, but subjective testing indicates it's not too bad the way
Irony works.

I'll continue to run with this for a while for testing purposes. While I've tested it on quite a few
diagnostic messages, there may be others that the current format doesn't work equally well for.

### DIFF
--- a/server/src/TUManager.cpp
+++ b/server/src/TUManager.cpp
@@ -48,9 +48,15 @@ TUManager::TUManager()
   // significantly slower.
   //
   // -- https://github.com/Sarcasm/irony-mode/issues/4
-  if (CINDEX_VERSION < 6) {
-    parseTUOptions_ &= ~CXTranslationUnit_PrecompiledPreamble;
-  }
+#if CINDEX_VERSION < 6
+  parseTUOptions_ &= ~CXTranslationUnit_PrecompiledPreamble;
+#endif
+
+  // Keep going even after fatal errors, or flycheck-irony will only display the
+  // first error.
+#if CINDEX_VERSION >= 34
+  parseTUOptions_ |= CXTranslationUnit_KeepGoing;
+#endif
 }
 
 TUManager::~TUManager() {
--- a/server/src/TUManager.cpp
+++ b/server/src/TUManager.cpp
@@ -11,6 +11,7 @@
 
 #include "TUManager.h"
 
+#include <algorithm>
 #include <iostream>
 
 TUManager::TUManager()
@@ -105,9 +106,14 @@ TUManager::parse(const std::string &filename,
     argv.push_back(CLANG_BUILTIN_HEADERS_DIR);
 #endif
 
-    for (auto &flag : flags) {
+    // Add spell checking; benefits diagnostics output.
+    if (std::find_if(flags.begin(), flags.end(),
+                     [] (const std::string& arg) { return arg == "-fspell-checking"; })
+        == flags.end())
+      argv.push_back("-fspell-checking");
+
+    for (auto &flag : flags)
       argv.push_back(flag.c_str());
-    }
 
     tu = clang_parseTranslationUnit(
         index_,
--- a/irony-diagnostics.el
+++ b/irony-diagnostics.el
@@ -44,6 +44,9 @@
 (defun irony-diagnostics-column (diagnostic)
   (nth 2 diagnostic))
 
+(defun irony-diagnostics-column-end (diagnostic)
+  (nth 3 diagnostic))
+
 (defun irony-diagnostics-severity (diagnostic)
   (nth 4 diagnostic))
 
--- a/server/src/Irony.cpp
+++ b/server/src/Irony.cpp
@@ -17,8 +17,12 @@
 #include <algorithm>
 #include <cassert>
 #include <iostream>
+#include <string>
+#include <vector>
 
-static std::string cxStringToStd(CXString cxString) {
+namespace {
+
+std::string cxStringToStd(CXString cxString) {
   std::string stdStr;
 
   if (const char *cstr = clang_getCString(cxString)) {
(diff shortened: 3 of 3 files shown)


server/src/TUManager.cpp
```diff
@@ -48,9 +48,15 @@ TUManager::TUManager()
   // significantly slower.
   //
   // -- https://github.com/Sarcasm/irony-mode/issues/4
-  if (CINDEX_VERSION < 6) {
-    parseTUOptions_ &= ~CXTranslationUnit_PrecompiledPreamble;
-  }
+#if CINDEX_VERSION < 6
```
Sarcasm: The previous version was intentional.
Generally speaking I prefer to avoid the preprocessor, because if there is a typo you don't know it unless you build in this configuration. `CXTranslationUnit_PrecompiledPreamble` has always existed to my knowledge it was just "broken" in an older versions.

Sarcasm:
A few comments.

I would prefer spell checking to be provided by Emacs directly just as the user may want `-Wdocumentation` or other functionality. I don't think there is an advantage in hardcoding it.

Regarding the fix-it hints, there was a discussion on flycheck about adding a generic fixer interface.
https://github.com/flycheck/flycheck/issues/530#issuecomment-65954007
I'm not sure it's going anywhere right now but I think the fix-it hints should not be merged with the diagnostics. Instead they should be first-class citizen of the diagnostic returned by irony-server. Then flycheck-irony can do some merging if it adds value. It may be long before there is a good interface to fix-it hints in Emacs, but if we have the fix-it hints at our disposal, it may be feasable to add a popup menu. Honestly I think a simple popup menu listing the fix it hints should not be too difficult, a bit like Eclipse Fix it hints: https://wiki.eclipse.org/FAQ_What_is_a_Quick_Fix%3F

What do you think?

There is definitely things to take from both of your PR. I understand we are not in a perfect world and we have to compromise but I want to avoid adding stuff that will need to be undone if the opportunity to do the thing the right way happen one day.

Sarcasm:
I merged the first change as https://github.com/Sarcasm/irony-mode/commit/84f9c9f0644036c754ee4724d094aa4fa246e17f

server/src/TUManager.cpp
itollefsen: Thanks for the clarification.

In the diagnostic code, the preprocessor check was used, so I thought that was the convention and changed the one place I found it to be different.  I'll leave it alone.

itollefsen:
Two things, and I'll start with the biggest.

I said I would continue to run with this for a while to see how it worked. I have, and it doesn't. Or rather, the notes and fix-it hints don't really work out as I had hoped.

Only a few notes adds context/clarifications. When they do, they are really useful, but that's unfortunately a minority.

When notes are on separate lines, especially above the actual diagnostic, it all just becomes messy because I, as the user, haven't read the actual error/warning yet, so the notes just appear out of context. Actually, they appear out of context even if I have read the error/warning because the link between them is lost and there can be a other diagnostic messages in between the two.

The fix-it hints are much the same. Often, they're just one character and don't really add any meaning.

I was thinking around the same lines as the Flycheck comment you linked, or indeed along the same lines as yourself and collect them as first-class citizens. I though about combining them with company-irony as completion candidates when at a fix-it location. However, all that was left for later - this was just about getting better diagnostics.

So I'll re-write the last commit. I want the header include guard note that appears on a separate line, but other than that one, I'll skip notes and warnings. You can get them from the compiler, so this is just one of those cases where we'll have to settle for less-than-perfect. Flycheck isn't a substitute for actually compiling the code after all.

My second comment is in regards to `-fspell-checking`. There are some fundamental differences between it and other compiler arguments such as `-Wdocumentation`. Primarily that Clang has it on by default, but libclang does not (for performance reasons). I found it confusing that I was getting the spelling hints when I compiled the code, but not from the diagnostic messages. And unlike the notes and fix-hit hints, these are actually really useful all the time.

It's easy to imagine this bogging down a IDE if the checking isn't done in the background. With `-fspell-checking` on, it could be that it must look for alternate names every time the user presses a key. However, I didn't notice any slowdown when using it with Irony. Presumably because Irony is running in the background.

Also, `-Wdocumentation` (and presumably others) would probably be part of the compilation flags the user has set (and in the compilation database). `-fspell-checking` is unlikely to be set (or even considered) in a similar fashion exactly because Clang has it on by default.

That said, I can take the same route and modify my compilation database to include `-fspell-checking` if you don't want it hardcoded.

Sarcasm:
Regarding `-spell-checking`, think it could be an option like that in elisp:

```el
(defcustom irony-libclang-spell-checking t
"Provide spell check messages.
Did you mean xxx?")

; modified version of `irony--adjust-compile-options`
(defun irony--adjust-compile-options ()
  (append
   (irony--lang-compile-option)
   (when irony-libclang-spell-checking
      '("-spell-checking"))
   ...))
```

Alternatively, it could just be in `irony-additional-clang-options`, user will be able to remove it if they 

I don't think it should come from the compilation database since you don't have to specify it on the clang command line.

itollefsen:
I'd like to drop this and come back to it.
I did the rewrite I mentioned I would do, and it works. I've been running it since then.
I'll post another PR on it at a later date. It probably needs a bit of updating at this point.

Sarcasm:
Ok, great, looking forward to it (totally forgot this PR ever existed but reading it again, it sounds interesting!). :)

PR #324, opened 2016-08-04
Define package local function string-suffix-p

kosh04:
string-suffix-p is new to Emacs 24.4

`emacs24` package on Travis-CI provides GNU Emacs 24.3.1


Sarcasm:
Thank you, pushed as 9f6d2f42a344dfe255e71c672dd7314db8de39d6!

PR #330, opened 2016-09-12
Don't quote compiler arguments.

Hylen:
This is sort of what the code does at the moment:
`(insert
 (combine-and-quote-strings '("-DMY_VAR=\"foo\"" "--std=c++11")))`
which becomes
`"-DMY_VAR=\"foo\"" --std=c++11`

This changes it to do
`(insert
 (mapconcat 'identity '("-DMY_VAR=\"foo\"" "--std=c++11") " "))`
becoming the desired
`-DMY_VAR="foo" --std=c++11`

### DIFF
--- a/irony.el
+++ b/irony.el
@@ -337,6 +337,10 @@ breaks with escaped quotes in compile_commands.json, such as in:
       (setq args (cons (apply 'string (nreverse cur-arg)) args)))
     (nreverse args)))
 
+(defun irony--combine-strings (strings)
+  "Join STRINGS with one space between every two elements."
+  (mapconcat 'identity strings " "))
+
 
 ;;
 ;; Mode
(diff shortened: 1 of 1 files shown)


Sarcasm:
The current behavior was done on purpose.

What is broken? Server-side, the function `unescapeCommandLine()` is used and is supposed to transform `"-DMY_VAR=\"foo\""` to `-DMY_VAR="foo"`. Did you notice that it does not work?

How do you handle spaces in compile options/filenames with your method? E.g: `-DMY_VAR="foo bar"`

Hylen:
> What is broken? Server-side, the function unescapeCommandLine() is used and is supposed to transform "-DMY_VAR=\"foo\"" to -DMY_VAR="foo". Did you notice that it does not work?

Yes, I saw one case when it didn't work for me. I tested the server using the command line and noticed that one argument was quoted. As I saw no reason why it should have been quoted I removed the quote manually and then it worked. It has been working for me since.

I have not actually tried this patch on your master, only on my branch with the compilation database changes. That branch doesn't change how the server parses the command line arguments though. Perhaps there is nothing wrong on master. I can unfortunately not upload any code or database files, as the project I found this in is proprietary. I will take a closer look when I have time.

Sarcasm:
Ok,
Let me know if you find the reason of the error, which compilation database is involved and what the input looked like.

Tack.

Hylen:
Hi again,

It turned out that this was an error on my branch after all. Sorry for the false alarm.

Merci beaucoup :)

Issue #333, opened 2016-09-16
irony-cdb-json--guess-flags returns wrong compile flags


stunpix:
I'm working with project based on Makefile and I'm using `bear` to generate `compile_commands.json`. At some point I opened one of my header files and one include was marked by `flycheck` as not found. Meanwhile my project is still able to build.

I started digging into irony sources and discovered strange behavior inside `irony-cdb-json.el`. I put debug messages in two functions:

```elisp
(defun irony-cdb-json--get-compile-options ()
  (irony--awhen (irony-cdb-json--locate-db)
    (let ((db (irony-cdb-json--load-db it)))
      (irony--aif (irony-cdb-json--exact-flags db)
          it
        (let ((dir-cdb (irony-cdb-json--compute-directory-cdb db)))
          ;; My debug message
          (let ((found-result (irony-cdb-json--guess-flags dir-cdb)))
            (message "json cdb get opts: %s" found-result)
            found-result))))))

(defun irony-cdb-json--guess-flags (dir-cdb)
  (cl-loop for e in dir-cdb
           with buf-path = (irony-cdb-json--target-path)
           with found = nil
           for dir = (car e)
           ;; keep the closest directory
           if (and (string-prefix-p dir buf-path)
                   (> (length dir) (length found)))
           do (setq found e)
           ;; My debug message
           (message "json cdb found: %s" (list (cons (nth 1 e) (nth 2 e))))
           finally return (list (cons (nth 1 e) (nth 2 e)))))
```

I killed all buffers with c/c++ sources and reopened header file with issue. Here is what I've got:

```
json cdb found: (((-pipe -O -W -Wall -Wpointer-arith -Wno-unused-parameter -Werror -g -I/home/alexander/prj/nginx-1.11.3/mods/nginx-rtstat/rtstatd/include -I src/core -I src/event -I src/event/modules -I src/os/unix -I objs -I src/http -I src/http/modules) . /home/alexander/prj/nginx-1.11.3))

json cdb get opts: (((-pipe -O -W -Wall -Wpointer-arith -Wno-unused-parameter -Werror -g -I/home/alexander/prj/nginx-1.11.3/mods/nginx-rtstat/rtstatd/include -I src/core -I src/event -I src/event/modules -I src/os/unix -I objs) . /home/alexander/prj/nginx-1.11.3))
```

[...]

PR #334, opened 2016-09-16
Fix issue #333 wrong compile flags returned

stunpix:
Function irony-cdb-json--guess-flags returns compile flags from last entry of compile_commands.json instead of best matched.


Sarcasm:
Thanks!

PR #336, opened 2016-09-28
checking environment variable to find libclang

little-big-h:


Sarcasm:
Where is the LLVM_ROOT environment variable used? I can't find a real reference to it in the LLVM source code.

little-big-h:
I don't know that it is used in LLVM but it seems a (loose) convention in cmake modules. Boost uses BOOST_ROOT, GTest uses GTEST_ROOT, dart uses DART_ROOT, ...

Sarcasm:
For now I prefer to recommend using `-DLIBCLANG_LIBRARY=<path>` and `-DLIBCLANG_INCLUDE_DIR=<path>`.

Ideally at some point we would have one variable to locate the LLVM root in one place but this variable would be the `llvm-config` executable path.

little-big-h:
*ideally*, you would use the cmake files that come with llvm

Sarcasm:
that's right ;)

greyblue9:
Oddly, even using

       -DCLANG_INSTALL_PREFIX=/usr/lib/llvm-3.8
       -DLibclang_LIBRARY=/usr/lib/llvm-3.8/lib/libclang-3.8.so.1
       -DLibclang_INCLUDE_DIR=/usr/lib/llvm-3.8/include

as mentioned above to force the use of LLVM/Clang 3.8, it still somehow ended up using LLVM 8 (with -I/usr/lib/llvm-8/include being the telltale sign) until I physically *broke* the LLVM8 install by renaming the directory with a trailing underscore, then it used my preferred version.

Trying the suggested patch now.

Any idea why this woulldn't work? I also had LLVM_ROOT set in my environment at the time (something I always do when configuring CMake projects).

Sarcasm:
Why prefer LLVM 3.8 to LLVM 8?
LLVM 8 is more recent, more featureful.
If you set `-DLibclang_INCLUDE_DIR=/usr/lib/llvm-3.8/include` explicitly, it's weird though, it should be the one that get picked up, disabling auto-detection.

greyblue9:
> Why prefer LLVM 3.8 to LLVM 8?
Only because I mistakenly thought that was the recommended version and didn't want two LLVM's taking up space. :)
It's working now though, thank you!

PR #359, opened 2017-02-11
Remove bundled ert.el

tarsius:
Ert was added to GNU Emacs in version 24.1, which was released in
2012.  Bundling a built-in or third-part library should generally
be avoided because the bundled version might end up being loaded
for users who also have a more recent version installed.


Sarcasm:
Thanks but some files that use this should be updated as well.
In practice it should not conflict as the file is only loaded by the test config when ert is not found but I agree it can be deleted.
Probably need to drop the pretended support for Emacs 23.x serie and remove this obselete package.

tarsius:
> Thanks but some files that use this should be updated as well.

I overlooked that, sorry. By the way, if you did want to keep supporting Emacs 23, then I would recommend adding the file `server/test/elisp/support/.nosearch`. That would make it much less likely that the bundled copy would end-up on the `load-path`.

Anyway, I just amended to my commit to remove the `load-path` wizardry at the end of `test-config.el`. I didn't find another place that needed to be adjusted.

Sarcasm:
I cherry-picked your commit as 284ca00b48d8261106efc6c6ac128ebec608c86f, thank you!

tarsius:
Thanks!

PR #368, opened 2017-03-18
Fixed compatibility with Gentoo's clang-4 installation directories.

nick-wallingford:
With llvm-4, Gentoo has a different directory versioning scheme. Index.h and libclang.so are located in:

     ~ $ find /usr/lib/llvm -name Index.h -or -name libclang.so
    /usr/lib/llvm/4/include/clang-c/Index.h
    /usr/lib/llvm/4/lib64/libclang.so

I don't know that this is the best way to solve this problem.

### DIFF
--- a/server/cmake/modules/FindLibClang.cmake
+++ b/server/cmake/modules/FindLibClang.cmake
@@ -16,7 +16,7 @@
 # most recent versions come first
 # http://llvm.org/apt/
 set(LIBCLANG_KNOWN_LLVM_VERSIONS 5.0.0 5.0
-  4.0.0 4.0
+  4.0.0 4.0 4
   3.9.1 3.9.0 3.9
   3.8.1 3.8.0 3.8
   3.7.1 3.7.0 3.7
@@ -46,6 +46,8 @@ foreach (version ${LIBCLANG_KNOWN_LLVM_VERSIONS})
     "/usr/local/lib/llvm-${version}/include"
     # FreeBSD ports versions
     "/usr/local/llvm${undotted_version}/include"
+    # Gentoo clang-4
+    "/usr/lib/llvm/${version}/include"
     )
 
   list(APPEND libclang_llvm_lib_search_paths
@@ -59,6 +61,9 @@ foreach (version ${LIBCLANG_KNOWN_LLVM_VERSIONS})
     "/usr/local/lib/llvm-${version}/lib"
     # FreeBSD ports versions
     "/usr/local/llvm${undotted_version}/lib"
+    # Gentoo clang-4
+    "/usr/lib/llvm/${version}/lib"
+    "/usr/lib/llvm/${version}/lib64"
     )
 endforeach()
 
--- a/server/cmake/modules/FindLibClang.cmake
+++ b/server/cmake/modules/FindLibClang.cmake
@@ -63,7 +63,6 @@ foreach (version ${LIBCLANG_KNOWN_LLVM_VERSIONS})
     "/usr/local/llvm${undotted_version}/lib"
     # Gentoo clang-4
     "/usr/lib/llvm/${version}/lib"
-    "/usr/lib/llvm/${version}/lib64"
     )
 endforeach()
 


server/cmake/modules/FindLibClang.cmake
```diff
@@ -59,6 +61,9 @@ foreach (version ${LIBCLANG_KNOWN_LLVM_VERSIONS})
     "/usr/local/lib/llvm-${version}/lib"
     # FreeBSD ports versions
     "/usr/local/llvm${undotted_version}/lib"
+    # Gentoo clang-4
+    "/usr/lib/llvm/${version}/lib"
+    "/usr/lib/llvm/${version}/lib64"
```
Sarcasm: Do you need to specify lib64? I think `cmake` handles it automatically, c.f:

> If the global property FIND_LIBRARY_USE_LIB64_PATHS is set all search paths will be tested as normal, with “64/” appended, and with all matches of “lib/” replaced with “lib64/”. This property is automatically set for the platforms that are known to need it if at least one of the languages supported by the PROJECT command is enabled.

-- https://cmake.org/cmake/help/v3.1/command/find_library.html?highlight=lib64

server/cmake/modules/FindLibClang.cmake
```diff
@@ -16,7 +16,7 @@
 # most recent versions come first
 # http://llvm.org/apt/
 set(LIBCLANG_KNOWN_LLVM_VERSIONS 5.0.0 5.0
```
Sarcasm: Before this line, could you try to add the following debug information and show me the output for my culture?

```cmake
find_package(Clang)

if (NOT Clang_FOUND)
   message(STATUS "not found: ${CLANG_INCLUDE_DIRS}")
else()
   if (NOT TARGET libclang)
     message(STATUS "libclang not a target???")
   endif()
   get_property(LIBCLANG_LIBRARY1
        TARGET libclang
        PROPERTY LOCATION)
   message(STATUS "Found: ${CLANG_INCLUDE_DIRS} ${LIBCLANG_LIBRARY1}")
endif()
```

Sarcasm:
Your fix looks somewhat reasonable to me, even though at some point I would like to rely on the Clang CMake configs to find these packages.

server/cmake/modules/FindLibClang.cmake
nick-wallingford: Indeed, the lib64 line can be left out.

server/cmake/modules/FindLibClang.cmake
nick-wallingford: -- C++11 compiler option(s): -std=c++11
    -- Found:  /usr/lib/llvm/4/lib64/libclang.so.4.0
    -- Found LibClang: /usr/lib64/llvm/4/lib64/libclang.so  
    -- Detecting libclang builtin headers directory
    -- Found:  /usr/lib/llvm/4/lib64/libclang.so.4.0
    -- Detecting libclang builtin headers directory -- success
    -- Irony package version is '0.2.2-cvs'

Looks like it's finding the library but not the headers?

Sarcasm: Could you show the content of `/usr/lib{,64}/cmake/clang/ClangConfig.cmake`? Not sure about the path, a more violent solution could be to use `find / -name ClangConfig.cmake` or something along these lines.

The content is not supposed to be too big.

nick-wallingford: path is `/usr/lib/llvm/4/lib64/cmake/clang/ClangConfig.cmake`

[ClangConfig.cmake.txt](https://github.com/Sarcasm/irony-mode/files/862702/ClangConfig.cmake.txt)

Sarcasm: I'm wondering why it is found then, this does not look like a standard CMake search path.
Ok thank you, it make sense, the variable was added quite recently (and was broken until yesterday: https://reviews.llvm.org/D30911).

We should be able to rely on this however for the next version of Clang.

Sarcasm:
Thank you for answering my questions and updating your PR, this is merged.

PR #369, opened 2017-03-24
Allow inaccessibles as candidates but emit a warning.

3246251196:
Allow inaccessibles as candidates but emit a warning.

### DIFF
--- a/irony-completion.el
+++ b/irony-completion.el
@@ -279,10 +279,11 @@ A candidate is composed of the following elements:
     Example: (\"(int a, int b)\" 1 6 8 13)
  7. The availability of the candidate."
   (when (irony-completion-candidates-available-p)
-    (cl-loop for candidate in irony-completion--candidates
-             when (member (irony-completion-availability candidate)
-                          irony-completion-availability-filter)
-             collect candidate)))
+    (cl-remove-if-not
+     (lambda (candidate)
+       (memq (irony-completion-availability candidate)
+             irony-completion-availability-filter))
+     irony-completion--candidates)))
 
 (defun irony-completion-candidates-async (callback)
   "Call CALLBACK when asynchronous completion is available.
--- a/irony-completion.el
+++ b/irony-completion.el
@@ -251,7 +251,6 @@ Return t if the context has been updated, nil otherwise."
 (defun irony-completion-post-comp-placeholders (candidate)
   (cdr (nth 7 candidate)))
 
-
 (defun irony-completion-candidates-available-p ()
   (and (eq (irony-completion--context-pos) irony-completion--context)
        (eq irony-completion--candidates-tick irony-completion--context-tick)))
@@ -269,15 +268,15 @@ A candidate is composed of the following elements:
  2. The [result-]type of the candidate, if any.
  3. If non-nil, contains the Doxygen brief documentation of the
     candidate.
- 4. The signature of the candidate excluding the result-type
+ 4. The accessiblity of the candidate.
+ 5. The signature of the candidate excluding the result-type
     which is available separately.
     Example: \"foo(int a, int b) const\"
- 5. The annotation start, a 0-based index in the prototype string.
- 6. Post-completion data. The text to insert followed by 0 or
+ 6. The annotation start, a 0-based index in the prototype string.
+ 7. Post-completion data. The text to insert followed by 0 or
     more indices. These indices work by pairs and describe ranges
     of placeholder text.
-    Example: (\"(int a, int b)\" 1 6 8 13)
- 7. The accessiblity of the candidate."
+    Example: (\"(int a, int b)\" 1 6 8 13)"
   (and (irony-completion-candidates-available-p)
        irony-completion--candidates))
 
--- a/irony-completion.el
+++ b/irony-completion.el
@@ -57,11 +57,11 @@
   :type '(repeat function)
   :group 'irony-completion)
 
-(defcustom irony-completion-accessibility-filter '(available deprecated)
- "For completion, only accept candidates whose accessibility is in the list.
+(defcustom irony-completion-availability-filter '(available deprecated)
+ "For completion, only accept candidates whose availability is in the list.
 
 Maps to libclang's CXAvailabilityKind:
-* https://clang.llvm.org/doxygen/group__CINDEX.html#ggada331ea0195e952c8f181ecf15e83d71af6542f250a1d21e256cd543669a49435
+- https://clang.llvm.org/doxygen/group__CINDEX.html#gada331ea0195e952c8f181ecf15e83d71
 
 Due to a bug in
 Clang (https://bugs.llvm.org//show_bug.cgi?id=24329), candidates
@@ -239,17 +239,18 @@ Return t if the context has been updated, nil otherwise."
 (defun irony-completion-brief (candidate)
   (nth 3 candidate))
 
-(defun irony-completion-accessibility (candidate)
-  (nth 4 candidate))
-
 (defun irony-completion-annotation (candidate)
-  (substring (nth 5 candidate) (nth 6 candidate)))
+  (substring (nth 4 candidate) (nth 5 candidate)))
 
 (defun irony-completion-post-comp-str (candidate)
-  (car (nth 7 candidate)))
+  (car (nth 6 candidate)))
 
 (defun irony-completion-post-comp-placeholders (candidate)
-  (cdr (nth 7 candidate)))
+  (cdr (nth 6 candidate)))
+
+(defun irony-completion-availability (candidate)
+  "See `irony-completion-availability-filter'"
+  (nth 7 candidate))
 
 (defun irony-completion-candidates-available-p ()
   (and (eq (irony-completion--context-pos) irony-completion--context)
@@ -268,17 +269,20 @@ A candidate is composed of the following elements:
  2. The [result-]type of the candidate, if any.
  3. If non-nil, contains the Doxygen brief documentation of the
     candidate.
- 4. The accessiblity of the candidate.
- 5. The signature of the candidate excluding the result-type
+ 4. The signature of the candidate excluding the result-type
     which is available separately.
     Example: \"foo(int a, int b) const\"
- 6. The annotation start, a 0-based index in the prototype string.
- 7. Post-completion data. The text to insert followed by 0 or
+ 5. The annotation start, a 0-based index in the prototype string.
+ 6. Post-completion data. The text to insert followed by 0 or
     more indices. These indices work by pairs and describe ranges
     of placeholder text.
-    Example: (\"(int a, int b)\" 1 6 8 13)"
-  (and (irony-completion-candidates-available-p)
-       irony-completion--candidates))
+    Example: (\"(int a, int b)\" 1 6 8 13)
+ 7. The availability of the candidate."
+  (when (irony-completion-candidates-available-p)
+    (cl-loop for candidate in irony-completion--candidates
+             when (member (irony-completion-availability candidate)
+                          irony-completion-availability-filter)
+             collect candidate)))
 
 (defun irony-completion-candidates-async (callback)
   "Call CALLBACK when asynchronous completion is available.
--- a/irony-completion.el
+++ b/irony-completion.el
@@ -57,6 +57,19 @@
   :type '(repeat function)
   :group 'irony-completion)
 
+(defcustom irony-completion-accessibility-filter '(available deprecated)
+ "For completion, only accept candidates whose accessibility is in the list.
+
+Maps to libclang's CXAvailabilityKind:
+* https://clang.llvm.org/doxygen/group__CINDEX.html#ggada331ea0195e952c8f181ecf15e83d71af6542f250a1d21e256cd543669a49435
+
+Due to a bug in
+Clang (https://bugs.llvm.org//show_bug.cgi?id=24329), candidates
+that can be validly accessed are deemed not-accessible."
+ :type '(repeat symbol)
+ :options '(available deprecated not-accessible)
+ :group 'irony-completion)
+
 
 ;;
 ;; Internal variables
@@ -226,17 +239,18 @@ Return t if the context has been updated, nil otherwise."
 (defun irony-completion-brief (candidate)
   (nth 3 candidate))
 
+(defun irony-completion-accessibility (candidate)
+  (nth 4 candidate))
+
 (defun irony-completion-annotation (candidate)
-  (substring (nth 4 candidate) (nth 5 candidate)))
+  (substring (nth 5 candidate) (nth 6 candidate)))
 
 (defun irony-completion-post-comp-str (candidate)
-  (car (nth 6 candidate)))
+  (car (nth 7 candidate)))
 
 (defun irony-completion-post-comp-placeholders (candidate)
-  (cdr (nth 6 candidate)))
+  (cdr (nth 7 candidate)))
 
-(defun irony-completion-inaccessible (candidate)
-  (nth 7 candidate))
 
 (defun irony-completion-candidates-available-p ()
   (and (eq (irony-completion--context-pos) irony-completion--context)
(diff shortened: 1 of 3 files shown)


server/src/Irony.cpp
```diff
@@ -346,8 +356,16 @@ void Irony::complete(const std::string &file,
         }
       }
 
+      // Raise a warning if the candidate may be susceptible to:
+      // https://github.com/Sarcasm/irony-mode/issues/367
+      if (availability == CXAvailability_NotAccessible) {
+        brief = "Irony: Warning: Candidate: `" +
+                prototype +
+                "' maybe be inaccessible due to protected or private access."
+                "\n";
```
Sarcasm: This newline makes the minibuffer resizing even when not necessary (i.e. there is no brief).

server/src/Irony.cpp
```diff
@@ -346,8 +356,16 @@ void Irony::complete(const std::string &file,
         }
       }
 
+      // Raise a warning if the candidate may be susceptible to:
+      // https://github.com/Sarcasm/irony-mode/issues/367
+      if (availability == CXAvailability_NotAccessible) {
+        brief = "Irony: Warning: Candidate: `" +
+                prototype +
```
Sarcasm: Showing the prototype does not seem useful, we don't show it for the brief, the user is supposed to already be aware of the context.

server/src/Irony.cpp
```diff
@@ -346,8 +356,16 @@ void Irony::complete(const std::string &file,
...
+                "' maybe be inaccessible due to protected or private access."
```
Sarcasm: maybe be => may be

Sarcasm:
After trying it out, I feel like this would be really nice if it was possible to opt-in/opt-out.
I find this too noisy to have the private member variables being shown.

The best behavior I think would be to have 2 backends:

1. the actual backend
2. another backend with only the accessible candidates

Then company-mode user could face 3 choices:

1. Do nothing, the behavior stays the same as now
2. Configure `company-backends` to insert the 2 backends one after the other:
    ```el
    (eval-after-load 'company
      '(progn
         (add-to-list 'company-backends 'company-irony-other)
         (add-to-list 'company-backends 'company-irony)))
    ```
    Then, one just need to bind `company-other-backend` for the rare occasion it needs one of the special candidates.

3. Configure `company-backends` to merge the 2 backends:
    ```el
    (eval-after-load 'company
      '(add-to-list 'company-backends '(company-irony company-irony-other)))
    ```

Personally I would be a user of option 2.

server/src/Irony.cpp
3246251196: Agreed overall. I found this useful because it actually changed the size of the mini-buffer which made me look down there. Should I apply a new line only if brief is appended to further down below ?

server/src/Irony.cpp
3246251196: Agreed.

server/src/Irony.cpp
3246251196: Agreed.

3246251196:
I will look into applying option 2.

server/src/Irony.cpp
Sarcasm: Yeah, that's what I did locally and I liked it better.

3246251196:
@Sarcasm 
Are you opposed to converting the prototype (etc) std::string to stringstreams, and getting const receiving references to rvalues whenever possible from the clang APIs rather than copies ?

Sarcasm:
I'm slightly opposed for the following reasons:

- will take time to review
- I would consider the issues with irony-mode are mostly architectural at this point
- can you measure the improvement of stringstream vs string? maybe an even better fix is string::reserve() + string::append()?

If you want to preserve some memory, I limit the number of translation units in the TUManager, because irony-server memory usage only grows over time because of this.

FYI, in the past the functions were writing std::string, then directly writing to std::ostream (std::cout), then back to std::string.
I don't think performances or memory usage have ever been impacted by this, there are bigger issues at play here.

If I wanted to optimize things like completion, I would do the filtering server-side instead of returning all the results to elisp and then filter.
I would not convert the CXString to std::string just to dump them.

TBH, my biggest grief is with the architecture of the Elisp code and irony-server.
This is quite limiting in extending and improving the functionality of irony-mode.
If I had to rewrite irony-server, I would probably try to follow the Language Server Protocol now that this sort of things exists. But right now, I'm waiting to see what the Clang's folks are going to achieve with clangd: https://github.com/llvm-mirror/clang-tools-extra/tree/master/clangd

Sarcasm:
To finish on a better note.
I'm not opposed to code improvements but the truth is I'm not sure stringstream is better than std::string.

3246251196:
+= is more efficient which I never realised until testing. And, more efficient by a long shot despite what colleagues at work say.

This PR will be specifically for the feature anyway. It will take me a while to do the el side of this but it is good practice.

Merci.

3246251196:
@Sarcasm 
Is it not possible to add a defcustom (irony-allow-inaccessbiles) which defaults to false. Then, use this in the filter function to not append candidates whose inaccessible flag is t if allow-ina* is set to f ?

Sarcasm:
It could be fine.
I think I could implement my other backend by let-binding this variable.

irony.el
```diff
@@ -99,6 +99,19 @@ automatically buffer-local wherever it is set."
   "C/C++ minor mode powered by libclang."
   :group 'c)
 
+(defcustom irony-allow-inaccessibles nil
```
Sarcasm: This variable does not seemed used at the moment, did you forget to push some of your code?

Anyway, now I feel like it would be better to just return the availability because it might be interesting to distinguish deprecated candidate as well, and this even once the Clang bug has been resolved.

If the accessibility is exposed to Elisp, then the defcustom variable could be something like:

```el
(defcustom irony-completion-accessibility-filter '(available deprecated)
 "For completion, only accept candidates whose accessibility is in the list.

Maps to libclang's CXAvailabilityKind:
* https://clang.llvm.org/doxygen/group__CINDEX.html#ggada331ea0195e952c8f181ecf15e83d71af6542f250a1d21e256cd543669a49435

Due to a bug in Clang (BUG REFERENCE HERE), it may be interesting to enable non-accessible candidates,
blabla."
 :type '(repeat symbol)
 :options '(available deprecated not-available)
 :group 'irony-completion)
```

I think not-accessible can be left out for now and can be added later if a use if found.
By the way,  this should be moved to `irony-completion.el`.

server/src/Irony.cpp
```diff
@@ -258,17 +259,30 @@ void Irony::complete(const std::string &file,
...
       typedtext.clear();
       brief.clear();
       resultType.clear();
       prototype.clear();
       postCompCar.clear();
       postCompCdr.clear();
+      inaccess.clear();
+
+      // progress even if candidate is `CXAvailability_NotAccessible' due to the
+      // current Clang bug which evaluates a protected member to be inaccessible
```
Sarcasm: Please reference the bug here, so that one can check it when reading this code and maybe update the code once the bug is fixed.

server/src/Irony.cpp
```diff
@@ -258,17 +259,30 @@ void Irony::complete(const std::string &file,
...
+      // to a derived class which has valid access.
+      //
+      // The rationale vs previous behaviour (which would not show an
```
Sarcasm: I would remove this paragraph.

server/src/Irony.cpp
```diff
@@ -258,17 +259,30 @@ void Irony::complete(const std::string &file,
...
+      // inaccessible candidate) is that the user's compiler catches erroneous
+      // access anyway. Since this is a completion engine, showing more is
+      // better than less.
+      //
+      // https://github.com/Sarcasm/irony-mode/issues/367
+      if (availability == CXAvailability_NotAvailable) {
+        continue;
+      } else if (availability == CXAvailability_NotAccessible) {
+        inaccess = "t";
```
Sarcasm: You should always set inaccess to a value, otherwise if an e=8th fields is added things will get confusing.

irony.el
3246251196: Hello,

I did not use this variable because I was unsure how to use it in the filtering function.

I presume you mean that not-available candidates can be left out for now since a derived's protected member is deemed not-accessible, not not-available?

3246251196: Also, I take it you want to record the availability  when the candidate is anything other than not-available since in Irony.cpp we `continue' the loop to save unnecessary processing?

Sarcasm: It looks like you presumed and understood right.

irony-completion.el
```diff
@@ -57,6 +57,19 @@
   :type '(repeat function)
   :group 'irony-completion)
 
+(defcustom irony-completion-accessibility-filter '(available deprecated)
+ "For completion, only accept candidates whose accessibility is in the list.
+
+Maps to libclang's CXAvailabilityKind:
+* https://clang.llvm.org/doxygen/group__CINDEX.html#ggada331ea0195e952c8f181ecf15e83d71af6542f250a1d21e256cd543669a49435
```
Sarcasm: Apparently I provided the wrong link, the right one should point to the enum, not to the NotAccessible value:
- https://clang.llvm.org/doxygen/group__CINDEX.html#gada331ea0195e952c8f181ecf15e83d71

nit: I used a start '*' because Github did not format a dash '-' correctly, could you replace the "list item" with a dash?
This is what I use in other places.

irony-completion.el
```diff
@@ -252,11 +268,12 @@ A candidate is composed of the following elements:
  2. The [result-]type of the candidate, if any.
  3. If non-nil, contains the Doxygen brief documentation of the
     candidate.
- 4. The signature of the candidate excluding the result-type
+ 4. The accessiblity of the candidate.
```
Sarcasm: Any reason to put this to the 4th position?
The `irony-completion-candidates` function is publicly documented the candidate "objects" are supposed to be "ABI stable" and extensible (by adding new fields).

Nit: could add a reference to `irony-completion-accessibility-filter` which lists the possible values.

irony-completion.el
```diff
@@ -57,6 +57,19 @@
   :type '(repeat function)
   :group 'irony-completion)
 
+(defcustom irony-completion-accessibility-filter '(available deprecated)
```
Sarcasm: Just noticed, Clang calls this "AvailabilityKind", wouldn't be better to use availability in favor of accessibility?
For example, `deprecated` may fit better in availability than accessibility.

server/src/Irony.cpp
```diff
@@ -245,7 +245,8 @@ void Irony::complete(const std::string &file,
     std::cout << "(\n";
 
     // re-use the same buffers to avoid unnecessary allocations
-    std::string typedtext, brief, resultType, prototype, postCompCar;
+    std::string typedtext, brief, resultType, prototype, postCompCar, access;
```
Sarcasm: access -> availability

server/src/Irony.cpp
```diff
@@ -258,17 +259,28 @@ void Irony::complete(const std::string &file,
...
+
+      // https://github.com/Sarcasm/irony-mode/issues/367
+      switch (availability) {
+      case CXAvailability_Available:
+        access = "available";
+        break;
+      case CXAvailability_Deprecated:
+        access = "deprecated";
+        break;
+      case CXAvailability_NotAvailable:
```
Sarcasm: Can you put this special case first or last, with a newline to separate from the others?

irony-completion.el
Sarcasm: By the way, regarding this function `irony-completion-candidates`.
I think it should be the one doing the filtering, IMHO you don't need the patch for company-irony (https://github.com/Sarcasm/company-irony/pull/32/).

The reason is that, the config is actually in irony-completion and not company-irony so it make sense to do the filtering here.
Secondly, it should also work for `irony-completion-at-point`
And finally, it is still possible for `company-irony` to introduce custom filtering just by let-binding the availability before calling this function.

server/src/Irony.cpp
3246251196: availability is a var name already used. i have used available - but this does exceed the 80 char limit ... ill let you decide

server/src/Irony.cpp
3246251196: i think i have done what you asked for. i have assumed you mean the special case to be adding not-accessible and added this to the bottom with a newline and the reference to the protected bug.

irony-completion.el
```diff
@@ -259,9 +276,13 @@ A candidate is composed of the following elements:
...
     more indices. These indices work by pairs and describe ranges
     of placeholder text.
-    Example: (\"(int a, int b)\" 1 6 8 13)"
-  (and (irony-completion-candidates-available-p)
-       irony-completion--candidates))
+    Example: (\"(int a, int b)\" 1 6 8 13)
+ 7. The availability of the candidate."
+  (when (irony-completion-candidates-available-p)
+    (cl-loop for candidate in irony-completion--candidates
+             when (member (irony-completion-availability candidate)
```
Sarcasm: Replace `member` by `memq` since we are dealing with symbols.

The `cl-loop` is doing a filter operation, so I would use the more descriptive `cl-remove-if-not`.

irony-completion.el
3246251196: i have attempted to do this. i am not sure about the elisp implementation at all, but it seems to work for me so far.

the only issue i have with this is that now we are kind of having n^2 since we loop through the candidates in irony-completion-candidates and in the company backend.. but perhaps this is negligible given the size of the data to process...

Sarcasm: It could be non-negligible, I could not say easily.
If the list of candidates is quite big I think most of the time is spent in irony-server and in the process filter.

server/src/Irony.cpp
Sarcasm: I think it is actually exactly 80 characters so it should be ok.
I'm ok with available.

server/src/Irony.cpp
Sarcasm: Sorry for the confusion, I meant the special case of "continuing", not exposing not-available to elisp.

server/src/Irony.cpp
```diff
@@ -258,17 +259,29 @@ void Irony::complete(const std::string &file,
...
-
       typedtext.clear();
       brief.clear();
       resultType.clear();
       prototype.clear();
       postCompCar.clear();
       postCompCdr.clear();
+      available.clear();
+
+      switch (availability) {
```
Sarcasm: For me the special case here is the case don't transform to string and ignore.
I would put it first and add a small comment:

      case CXAvailability_NotAvailable:
          // no benefits to expose this to Elisp for now
          continue;

server/src/Irony.cpp
```diff
@@ -258,17 +259,29 @@ void Irony::complete(const std::string &file,
...
+      case CXAvailability_Available:
+        available = "available";
+        break;
+      case CXAvailability_Deprecated:
+        available = "deprecated";
+        break;
+      case CXAvailability_NotAvailable:
+        continue;
+
+      // https://github.com/Sarcasm/irony-mode/issues/367
```
Sarcasm: Remove this link, I don't think it helps understand the code.

irony-completion.el
3246251196: I was wrong about n^2 anyway since this would result in 2n which is nothing!

Sarcasm:
Thank you very much, this change is much appreciated!

3246251196:
@Sarcasm 
Is it still useful to flag that the in-accessible candidate is risky in the mini buffer/ brief ? 

I think I will insert a hack so I get that functionality on my irony-server. How do you feel?

Sarcasm:
I feel like this can be better implemented in `company-irony`. For example in the help of deprecated candidate we could annotate `[deprecated]`, or many other ways to denote this.
The advantage to do it on the Elisp side is that this kind of information can also be highlighted for example (bold, grey, yellow warning, ...). Or maybe it could impact the candidate inline for example, by adding some sort of symbol to the item in the completion list.

3246251196:
Yeah I really preferred that idea. Of having a different highlighting colour for inaccessibles or whatever.

Maybe I will look into this.

PR #370, opened 2017-03-28
irony-completion.el (irony-completion-at-point): Filter based on prefix

abo-abo:
For example, completing "foo.bar|", the candidates should all match "bar".

### DIFF
--- a/irony-completion.el
+++ b/irony-completion.el
@@ -341,13 +341,21 @@ Note that:
 ;;;###autoload
 (defun irony-completion-at-point ()
   (when (and irony-mode (irony-completion-candidates-available-p))
-    (let ((symbol-bounds (irony-completion-symbol-bounds)))
+    (let* ((symbol-bounds (irony-completion-symbol-bounds))
+           (prefix (buffer-substring-no-properties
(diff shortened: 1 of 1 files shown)


Sarcasm:
Could you describe the issue and give a minimal working example?
Mine works fine without filtering and reading the doc quickly I could not convince myself this was or was not required.
Could you point me to the right place? Because I can imagine the completion-at-point function trying to be smart and have some sort of fuzzy complete, in this case it would make sense to return all the candidates.

```cpp

struct Foo {
  int barry;
  int barack;
  int obama;
};

void f(const Foo &foo) {
  foo.bar|                           # I get "barry" and "barack"
}
```

abo-abo:
I try to reproduce it now, but I can't. Yesterday, I got all completions in `cv` ignoring the prefix:

```C++
#include <opencv2/opencv.hpp>
cv::acc
```

abo-abo:
Closing for now; I'll come back if I manage to reproduce something:)

Sarcasm:
Ok great,
yes let me know if that's the case, I'm not opposed to the change if it is the right change. :)

Issue #358, opened 2017-02-05
Can company-irony cache its completion results?


dmalyuta:
In my current project (which is not big at all, only about 800 lines but includes heavy headers from Boost and PCL as well as some ROS headers) completion at point with company-irony can take up to 10 seconds. I'm fine if this happens the first time that I invoke a new completion, but the problem is that I may attempt to do a completion at one point and even just 5 seconds later, doing the exact same completion takes again up to 10 seconds! I would have thought that there is some sort of caching - i.e. if I completed a symbol once, it is kept in memory for some time so that similar completions can happen quickly. Is there any option for this?

I am using Emacs 25.1 on Ubuntu 14.04. Here is the part of my `init.el` relevant to completion:

```
(use-package company
    ;; complete anything
    :ensure t
    :bind
    (("<S-SPC>" . company-complete)
     ("C-x C-n" . company-select-next)
     ("C-x C-p" . company-select-previous))
    :init
    ;;(add-hook 'after-init-hook 'global-company-mode)
    (add-hook 'c-mode-common-hook 'company-mode)
    (add-hook 'emacs-lisp-mode-hook 'company-mode)
    :config
    (setq company-backends (delete 'company-semantic company-backends))
    ;; use C++11 by default
    (require 'cc-mode)
    (set 'company-clang-arguments (list "-std=c++11")))

  (use-package irony
    ;; asynchronous capabilities (improces C/C++ editing experience)
    ;; Linux Mint:
    ;;  sudo apt-get install g++ clang libclang-dev
    ;;  M-x irony-install-server # when in c-mode or c++-mode
    :ensure t
    :defer t
    :init
    (add-hook 'c-mode-common-hook 'irony-mode)
    :config
    ;; replace completion-at-point and complete-symbol by irony-mode's asynchronous functions
    (defun my-irony-mode-hook ()
      (define-key irony-mode-map [remap completion-at-point]
	'irony-completion-at-point-async)
      (define-key irony-mode-map [remap complete-symbol]
	'irony-completion-at-point-async))
    (add-hook 'irony-mode-hook 'my-irony-mode-hook)
    (add-hook 'irony-mode-hook 'irony-cdb-autosetup-compile-options)
    (setq irony-additional-clang-options '(;; use C++11
					   "-std=c++11"))
    ;; additional include paths
    ;;					 "-I/home/danylo/catkin_ws/devel/include/"
    ;;					 "-I/opt/ros/kinetic/include/"))
[...]

PR #382, opened 2017-05-02
Re-instate sentinal

3246251196:
This may seem like an obscure request, but it is actually due to: #358.

I concede that I have not yet tracked down issue 358 (and I will look into this the next chance I get!). Stopping irony-mode and restarting is the only way to remove obscure references that appear after the process has been active for a while (though this does not happen to me often); anyway, this issue is not about 358. I would like to request the patch above because when I kill irony I like to see the mini-buffer text.

### DIFF
--- a/irony.el
+++ b/irony.el
@@ -620,6 +620,7 @@ list (and undo information is not kept).")
                     (format-time-string "irony.%Y-%m-%d_%Hh-%Mm-%Ss.log")
                     temporary-file-directory))))
     (set-process-query-on-exit-flag process nil)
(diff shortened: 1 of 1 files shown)


Sarcasm:
Sorry, I pushed a fix after seeing your issue and before seeing the PR.

See https://github.com/Sarcasm/irony-mode/commit/349d007d55e06094a2223c14bcbc0025904e77b1

3246251196:
@Sarcasm 
Okay, cheers.

PR #388, opened 2017-05-23
Find server binary via locate-file

drvink:
This allows the binary to be found with or without an .exe extension.

### DIFF
--- a/irony.el
+++ b/irony.el
@@ -557,7 +557,9 @@ The installation requires CMake and the libclang developpement package."
 
 Throw an `irony-server-error' if a proper executable cannot be
 found."
-  (let ((exe (expand-file-name "bin/irony-server" irony-server-install-prefix)))
+  (let* ((exec-path (cons (expand-file-name "bin" irony-server-install-prefix)
+                          exec-path))
+         (exe (executable-find "irony-server")))
     (condition-case err
         (let ((version (car (process-lines exe "--version"))))
           (if (and (string-match "^irony-server version " version)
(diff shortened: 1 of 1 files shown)


Sarcasm:
Did you get an issue with not finding the exe?

Users like @mambolevis and @3246251196 never reported this, do you know what you do that could be different?

I'm feel like it would be better to reuse `executable-find`, what do you think?

Rough idea is something like:

```el
(let ((exec-path (cons (expand-file-name "bin" irony-server-install-prefix)
                  exec-path)))
    (executable-find "irony-server"))
```

drvink:
I initially did use executable-find in the exact way you mentioned, but for some reason, it didn't work in irony.el, even though it worked in an ielm session. I'm not sure why--I'd have expected that dynamic binding to work as well. (Maybe I did something wrong; I'm not at my computer now, so I can't check again--at any rate, the version using locate-file does work.)

Sarcasm:
Ok, would be nice to know if it really does not work.
And if it doesn't, at least reuse `exec-suffixes`.

drvink:
I tried it again and this time it worked, so I must have messed something up the first time. I re-pushed with a version using `executable-find`.

Sarcasm:
Great,
now it would be nice to fix the function so the tests pass.
The functions says:

> Throw an `irony-server-error' if a proper executable cannot be found."

You could leave the process-lines call as it is, and replace the 2 other references of `exe` by `(executable-find "irony-server")`, or something along these lines.

drvink:
Fixed; sorry about that.

irony.el
```diff
@@ -575,7 +577,8 @@ found."
        (signal (car err) (cdr err)))
       (error
        (signal 'irony-server-error
-               (if (file-executable-p exe)
+               (if (and (stringp exe)
```
Sarcasm: I prefer to check for `nil`, which is the sentinel value we can expect from `executable-find`:

```el
(if (and exe (file-executable-p exe))
```

drvink: changed

Sarcasm:
Great,
thank you for the patch and your patience with the review.

Issue #403, opened 2017-06-28
irony for json cdb don't support "arguments" key in compile_commands.json


ANDRON94:
Hi, I have the next json compilation database file: 

> [ 
    {
        "arguments": [
            "c++", 
            "-c", 
            "-DHAVE_CONFIG_H", 
            "-I.", 
            "-I../../src", 
            "-I..", 
            "-pthread", 
            "-I/usr/include", 
            "-pthread",
            "OTHER STUFF"
        ], 
        "directory": "/home/XXXXXX", 
        "file": "./mgmt/XXXXX.cpp"
    }
]

And it  didn't properly loaded into Emacs. I've got next error message:

> Irony CDB: error in compilation database: (irony-server-error cannot-load-database "failed to load compilation database from directory" "/home/XXXXX").

I inspect an Irony code a little bit(irony-cdb-json--compile-command-options) and if I don't miss something the current implementation only support "command" key not "arguments". But  from [JSON Compilation Database specification](http://clang.llvm.org/docs/JSONCompilationDatabase.html):

> *arguments*: The compile command executed as list of strings. Either arguments or command is required.

Sarcasm:
Indeed, this feature has been added somewhat recently to the compilation database specification.
It could work with the libclang backend if your clang version is up-to-date.
The JSON parsing done in Elisp needs to be updated, it should not be too difficult to contribute ;)

ANDRON94:
Yeah, I'll create a pull request asap.

PR #404, opened 2017-06-29
irony-cdb-json: fix parsing of compile command options

ANDRON94:
Adds support for "arguments" key in JSON compilation database. #403


ANDRON94:
Unfortunately I can't test that JSON database with "arguments" is properly loaded because I have the old version of libclang. But I test function itself and it's returns a list of commands, just like before.

Sarcasm:
Forgot about this PR, sorry about that.
It has been merged now.

ANDRON94:
No problems, I am glad I was able to help =)

PR #412, opened 2017-07-30
Add prefix filter for 'candidates' commands

kangjianbin:
This commit adds prefix parameter for candidates.
In irony-server, it will filter the completion results before sending them to client.
Prefix "*" is treated as no filtering (return all completions results) as I can't find an easy way to support optional parameter in irony-server

We still need modify company-irony to support. Since it is in another respository, the fix doesn't contain the modification.

### DIFF
--- a/irony-completion.el
+++ b/irony-completion.el
@@ -185,14 +185,15 @@ that can be validly accessed are deemed not-accessible."
 
 (irony-iotask-define-task irony--t-candidates
   "`candidates' server command."
-  :start (lambda (prefix)
-           (irony--server-send-command "candidates" prefix))
+  :start (lambda (prefix ignore-case)
+           (irony--server-send-command "candidates" prefix
+                                       (if ignore-case "on" "off")))
   :update irony--server-query-update)
 
-(defun irony--candidates-task (&optional buffer pos prefix)
+(defun irony--candidates-task (&optional buffer pos prefix ignore-case)
   (irony-iotask-chain
    (irony--complete-task buffer pos)
-   (irony-iotask-package-task irony--t-candidates prefix)))
+   (irony-iotask-package-task irony--t-candidates prefix ignore-case)))
 
 
 ;;
@@ -231,12 +232,7 @@ that can be validly accessed are deemed not-accessible."
            irony-completion-availability-filter))
    candidates))
 
-(defun irony-completion-make-prefix (prefix)
-  (if (and prefix (not (string= prefix "")))
-      prefix
-    "*"))
-
-(defun irony-completion-candidates (&optional prefix)
+(defun irony-completion-candidates (&optional prefix ignore-case)
   "Return the list of candidates at point.
 
 A candidate is composed of the following elements:
@@ -258,15 +254,13 @@ A candidate is composed of the following elements:
   (irony--awhen (irony-completion-symbol-bounds)
     (irony-completion--filter-candidates
      (irony--run-task
-      (irony--candidates-task nil (car it)
-                              (irony-completion-make-prefix))))))
+      (irony--candidates-task nil (car it) prefix ignore-case)))))
 
-(defun irony-completion-candidates-async (callback &optional prefix)
+(defun irony-completion-candidates-async (callback &optional prefix ignore-case)
   (irony--aif (irony-completion-symbol-bounds)
       (lexical-let ((cb callback))
         (irony--run-task-asynchronously
-         (irony--candidates-task nil (car it)
-                                 (irony-completion-make-prefix prefix))
+         (irony--candidates-task nil (car it) prefix ignore-case)
          (lambda (candidates-result)
            (funcall cb (irony-completion--filter-candidates
                         (irony-iotask-result-get candidates-result))))))
@@ -396,7 +390,13 @@ A candidate is composed of the following elements:
 (defun irony-completion-at-point ()
   (irony--awhen (and irony-mode (irony-completion-symbol-bounds))
     (let ((candidates (irony-completion--filter-candidates
-                       (irony--run-task (irony--candidates-task nil (car it))))))
+                       (irony--run-task
+                        (irony--candidates-task
+                         nil
+                         (car it)
+                         (buffer-substring-no-properties
+                          (car it) (cdr it))
+                         completion-ignore-case)))))
       (list
        (car it)                           ;start
        (cdr it)                           ;end
--- a/irony-completion.el
+++ b/irony-completion.el
@@ -185,19 +185,19 @@ that can be validly accessed are deemed not-accessible."
 
 (irony-iotask-define-task irony--t-candidates
   "`candidates' server command."
-  :start (lambda (prefix ignore-case)
+  :start (lambda (prefix style)
            (irony--server-send-command
             "candidates" prefix
-            (cond
-             ((not ignore-case) "")
-             ((eq ignore-case 'smart) "smart")
-             (t "insensitive"))))
+            (pcase style
+              ('case-insensitive "case-insensitive")
+              ('smart-case "smart-case")
+              (other "exact"))))
   :update irony--server-query-update)
 
-(defun irony--candidates-task (&optional buffer pos prefix ignore-case)
+(defun irony--candidates-task (&optional buffer pos prefix style)
   (irony-iotask-chain
    (irony--complete-task buffer pos)
-   (irony-iotask-package-task irony--t-candidates prefix ignore-case)))
+   (irony-iotask-package-task irony--t-candidates prefix style)))
 
 
 ;;
@@ -236,7 +236,7 @@ that can be validly accessed are deemed not-accessible."
            irony-completion-availability-filter))
    candidates))
 
-(defun irony-completion-candidates (&optional prefix ignore-case)
+(defun irony-completion-candidates (&optional prefix style)
   "Return the list of candidates at point.
 
 A candidate is composed of the following elements:
@@ -258,13 +258,13 @@ A candidate is composed of the following elements:
   (irony--awhen (irony-completion-symbol-bounds)
     (irony-completion--filter-candidates
      (irony--run-task
-      (irony--candidates-task nil (car it) prefix ignore-case)))))
+      (irony--candidates-task nil (car it) prefix style)))))
 
-(defun irony-completion-candidates-async (callback &optional prefix ignore-case)
+(defun irony-completion-candidates-async (callback &optional prefix style)
   (irony--aif (irony-completion-symbol-bounds)
       (lexical-let ((cb callback))
         (irony--run-task-asynchronously
-         (irony--candidates-task nil (car it) prefix ignore-case)
+         (irony--candidates-task nil (car it) prefix style)
          (lambda (candidates-result)
            (funcall cb (irony-completion--filter-candidates
                         (irony-iotask-result-get candidates-result))))))
@@ -400,7 +400,9 @@ A candidate is composed of the following elements:
                          (car it)
                          (buffer-substring-no-properties
                           (car it) (cdr it))
-                         completion-ignore-case)))))
+                         (if completion-ignore-case
+                             'case-insensitive
+                           'exact))))))
       (list
        (car it)                           ;start
        (cdr it)                           ;end
--- a/irony-completion.el
+++ b/irony-completion.el
@@ -185,14 +185,14 @@ that can be validly accessed are deemed not-accessible."
 
 (irony-iotask-define-task irony--t-candidates
   "`candidates' server command."
-  :start (lambda ()
-           (irony--server-send-command "candidates"))
+  :start (lambda (prefix)
+           (irony--server-send-command "candidates" prefix))
   :update irony--server-query-update)
 
-(defun irony--candidates-task (&optional buffer pos)
+(defun irony--candidates-task (&optional buffer pos prefix)
   (irony-iotask-chain
    (irony--complete-task buffer pos)
-   (irony-iotask-package-task irony--t-candidates)))
+   (irony-iotask-package-task irony--t-candidates prefix)))
 
 
 ;;
@@ -231,7 +231,12 @@ that can be validly accessed are deemed not-accessible."
            irony-completion-availability-filter))
    candidates))
 
-(defun irony-completion-candidates ()
+(defun irony-completion-make-prefix (prefix)
+  (if (and prefix (not (string= prefix "")))
+      prefix
+    "*"))
+
+(defun irony-completion-candidates (&optional prefix)
   "Return the list of candidates at point.
 
 A candidate is composed of the following elements:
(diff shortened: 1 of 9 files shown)


Sarcasm:
Great to see that you tackled this issue.
A few questions first, a quick look at the code looks ok but a few questions regarding the design choices:

- any reasons to modify the existing candidates functions over adding a new one? The way it is, it should not be an issue to have multiple functions (e.g. `prefix-candidates`, `fuzzy-candidates`).  It looks like the irony elisp API stays compatible so that may not be an issue.
- why can't we use an empty prefix ""? It would make sense to me, this prefix would not filter out any candidates, and this should not require any special case in the code.

It looks like `company-irony` needs needs to support case insensitive matching, so an additional command or something needs to be added to before integrating into company-irony.

Apart from that, looks like a great addition to me, I would be happy to test the whole thing, review the code and merge it.

kangjianbin:
>    any reasons to modify the existing candidates functions over adding a new one? The way it is, it should not be an issue to have multiple functions (e.g. prefix-candidates, fuzzy-candidates). It looks like the irony elisp API stays compatible so that may not be an issue.

In this implementation, 'candidates' command can cover 'prefix-candidates'. If we pass prefix as '*', 'candidates' behaviors as before; if pass valid prefix string, 'candidates' behaviors as 'prefix-candidates'.
There is no API change. But user need to recompile irony-server.

>    why can't we use an empty prefix ""? It would make sense to me, this prefix would not filter out any candidates, and this should not require any special case in the code.

I found that in function 
```elisp
(irony--server-send-command "candidates" prefix)
```
If I passed prefix as "", irony-server sees only 1 arguments ('candidates'). The problem is in function (combine-and-quote-strings):
```elisp
(combine-and-quote-strings '("a" "" "b"))
=> "a  b"
```

Sarcasm:
Ok, this is a pity `combine-and-quote-strings` has this behavior, I would prefer not to use the function and have no special-case for irony-server.
We can sort this out after working on company-irony though.

kangjianbin:
Hi Sarcasm,
I just sent another pull request in company-irony to support prefix parameter.
This is a small change since most work is done in irony-mode.

I did some test on a T460s laptop. For linux 4.3 kernel source completion, the candidates menu poped up in less than 0.5 seconds (length of prefix is 3).

kangjianbin:
Hi Sarcasm,
I just sent another pull request in company-irony to support prefix parameter.
This is a small change since most work is done in irony-mode.

I did some test on a T460s laptop. For linux 4.3 kernel source completion, the candidates menu poped up in less than 0.5 seconds (length of prefix is 3).

Sarcasm:
> I just sent another pull request in company-irony to support prefix parameter.

Great, I will take a look and test things on my side as well, thank you.

> For linux 4.3 kernel source completion, the candidates menu poped up in less than 0.5 seconds (length of prefix is 3).

And before that?

irony-completion.el
```diff
@@ -252,13 +257,16 @@ A candidate is composed of the following elements:
  7. The availability of the candidate."
   (irony--awhen (irony-completion-symbol-bounds)
     (irony-completion--filter-candidates
-     (irony--run-task (irony--candidates-task nil (car it))))))
+     (irony--run-task
+      (irony--candidates-task nil (car it)
+                              (irony-completion-make-prefix))))))
```
Sarcasm: This function call is missing the prefix argument, this triggers an error.

irony-completion.el
```diff
@@ -231,7 +231,12 @@ that can be validly accessed are deemed not-accessible."
            irony-completion-availability-filter))
    candidates))
 
-(defun irony-completion-candidates ()
+(defun irony-completion-make-prefix (prefix)
```
Sarcasm: I'm definitely not fond of this special case.
I would prefer to vendor `combine-and-quote-string` and make it handle the empty case.
I'm actually wondering if this should be reported to Emacs developers.

server/src/Command.cpp
```diff
@@ -204,7 +207,7 @@ Command *CommandParser::parse(const std::vector<std::string> &argv) {
     command_.flags.assign(std::next(argsEnd), argv.end());
   }
 
-  if (argCount != static_cast<int>(positionalArgs.size())) {
+  if (argCount < static_cast<int>(positionalArgs.size())) {
```
Sarcasm: Is this change still required? If not, I would prefer to revert it.

server/src/Irony.cpp
```diff
@@ -433,6 +433,13 @@ void Irony::candidates() const {
         typedTextSet = true;
       }
     }
+    if (prefix != "" && prefix != "*") {
+      auto res = std::mismatch(prefix.begin(), prefix.end(), typedtext.begin());
```
Sarcasm: Would be nice to make a function for that, something like `bool startsWith(const std::string &str, const std::string &prefix)`.

Also, could you make this check this earlier? In order to avoid the unnecessary processing that is going on to make the annotation, brief and all. In the if-block of line 429 for example: `if (chunkKind == CXCompletionChunk_TypedText && !typedTextSet)`.

kangjianbin:
Thanks for you review. I am glad to make the changes.

>this code needs to add support for case-insensitive prefix too, so that company-irony can make use of it.

What does that mean? I never use insensitive completion before. Looks like there is no insensitive matching in company irony or I miss something?

I plan to add a boolean parameter 'ignore_case' for candidates command:  candidates prefix ignore_case.
If ignore_case is true, irony-server will ignore case while matching prefix. Please let me know if this is incorrect.

server/src/Command.cpp
kangjianbin: No, I will revert it. I tried to support optional parameter, but failed.

Sarcasm:
There is the variable `company-irony-ignore-case`, that you can see here:
- https://github.com/Sarcasm/company-irony/blob/b8001ffc4da7403172835d47bc740cfbac35364d/company-irony.el#L45-L47

> I plan to add a boolean parameter 'ignore_case' for candidates command: candidates prefix ignore_case.
> If ignore_case is true, irony-server will ignore case while matching prefix. Please let me know if this is incorrect.

This sounds good to me. :+1:

kangjianbin:
Add support for ignore-case and some bugs fixed.

irony.el
```diff
@@ -645,8 +645,21 @@ list (and undo information is not kept).")
 (defun irony--run-task-asynchronously (task callback)
   (irony-iotask-schedule (irony--get-server-process-create) task callback))
 
+(defun irony--quote-strings (strings &optional separator)
```
Sarcasm: Just for future reference, could you comment that the implementation improves over `combine-and-quote-strings`?

kangjianbin:
The smart style looks good. I will implement it.

kangjianbin:
I found that smart case style doesn't work on completion-at-point as completion-ignore-case is a boolean variable.

server/src/Irony.cpp
```diff
@@ -286,6 +286,50 @@ void Irony::complete(const std::string &file,
...
+    : [] (char a, char b) { return std::tolower(a) == std::tolower(b); };
+
+  auto res = std::mismatch(prefix.begin(), prefix.end(), str.begin(), charCmp);
+  if (res.first != prefix.end()) {
+    return false;
+  }
+  return true;
+}
+
+static bool isCaseSensitive(const std::string &prefix, std::string caseStyle)
```
Sarcasm: Please use const-ref for caseStyle: `const std::string &caseStyle`.

server/src/Irony.cpp
```diff
@@ -286,6 +286,50 @@ void Irony::complete(const std::string &file,
   std::cout << "(success . t)\n";
 }
 
+namespace {
+
+static bool hasUppercase(const std::string &prefix)
```
Sarcasm: You can remove `static` here and the other functions below, I prefer unnamed namespaces to static function.
This is mostly a matter of style and not a real issue, see also https://stackoverflow.com/a/4977525/951426

server/src/Irony.cpp
```diff
@@ -286,6 +286,50 @@ void Irony::complete(const std::string &file,
...
+{
+  for (auto it = prefix.begin(); it != prefix.end(); ++it) {
```
Sarcasm: Is it std::find_if(b, e, std::isupper) ?

Something like:

    return std::find_if(prefix.begin(), prefix.end(), std::isupper) != prefix.end();

Not sure it compiles, just a suggestion.

server/src/Irony.cpp
```diff
@@ -286,6 +286,50 @@ void Irony::complete(const std::string &file,
...
+  if (str.length() < prefix.length()) {
+    return false;
+  }
+
+  auto charCmp = caseSensitive
+    ? [] (char a, char b) { return a == b; }
+    : [] (char a, char b) { return std::tolower(a) == std::tolower(b); };
+
+  auto res = std::mismatch(prefix.begin(), prefix.end(), str.begin(), charCmp);
+  if (res.first != prefix.end()) {
```
Sarcasm: the 4 next line are just `return res.first == prefix.end()`

server/src/Irony.cpp
```diff
@@ -286,6 +286,50 @@ void Irony::complete(const std::string &file,
...
+    return false;
+  }
+  return true;
+}
+
+static bool isCaseSensitive(const std::string &prefix, std::string caseStyle)
+{
+  if (caseStyle == "" || caseStyle == "nil") {
```
Sarcasm: I think it would make more sense to check "is caseStyle case-insensitive? Yes, then return true.", and default to case sensitive (function ends with `return false`).

server/src/Irony.h
```diff
@@ -100,7 +100,7 @@ class Irony {
   /// Get all the completion candidates.
   ///
   /// \pre complete() was called.
-  void candidates() const;
+  void candidates(const std::string &prefix, const std::string caseStyle) const;
```
Sarcasm: Use const-ref, not just const for caseStyle.

server/src/main.cpp
```diff
@@ -102,6 +102,7 @@ struct RestoreClogOnExit {
   std::streambuf *rdbuf_;
 };
 
+
```
Sarcasm: unnecessary?

server/src/Commands.def
```diff
@@ -12,7 +12,9 @@
 #error Please define the 'X(id, command, description)' macro before inclusion!
 #endif
 
-X(Candidates, "candidates", "print completion candidates (require previous complete)")
+X(Candidates, "candidates",
+  "PREFIX STYLE - print completion candidates (require previous complete). "
```
Sarcasm: I think, for irony.el/irony-completion.el and irony-server, I would prefer that you use consistently the term "style" for this candidate argument.

The style should be a list of the following values:

- in Elisp: `'exact`, `'case-insensitive` and `'smart-case`
- in C++ use an enum class for std::string caseStyle, so you benefit more from the type system, only when read/validate the command you deal with a string

I prefer the term style, because in the future I can see something not related to case, which would be fuzzy search.
It could be a new function, but if we have an arbitrary style value here, it is sufficient.

On the other hand, you change in `company-irony-ignore-case` is perfect, very sensible choice to keep compatibility,
but I think irony-server sides this could be a bit clearer and leaves the door open for the future.

kangjianbin: Sounds reasonable. 

> in Elisp: 'exact, 'case-insensitive and 'smart-case

So need we create a new variable in elisp or just reuse company-irony-ignore-case?

> in C++ use an enum class for std::string caseStyle, so you benefit more from the type system, only when read/validate the command you deal with a string

So we need create a new enum like 'enum Style' (or other name if you have better option :)) and a new StyleConverter to validate it. Both irony.cpp and command.cpp need access this enum. Is it ok to create a new header file for this enum?

Sarcasm: > So need we create a new variable in elisp or just reuse company-irony-ignore-case?

For company-irony company-irony-ignore-case is good.
For irony-completion, change the following signature:

```el
(defun irony-completion-candidates (&optional prefix ignore-case)
```

to:

```el
(defun irony-completion-candidates (&optional prefix style)
```


> So we need create a new enum like 'enum Style' (or other name if you have better option :)) and a new StyleConverter to validate it

- `enum class Style`
- `enum class PrefixStyle`
- `enum class MatchStyle`
- `enum class PrefixMatchStyle`
- `enum class PrefixMatchStyle`
- `enum class CandidatesStyle`

My favorite is `PrefixMatchStyle`, but you can pick up anyone you want, or one of your own if you find something better.

And yes a new converter, you are right.


> Both irony.cpp and command.cpp need access this enum. Is it ok to create a new header file for this enum?

New header is fine. If I had to pick an existing one, I would choose Irony.h, but new header is good to me too.

Sarcasm:
Hey, just for your information.
I did not look at the code yet, but testing it I already like `(setq company-irony-ignore-case 'smart)` very much, great work! :+1:

server/src/Commands.def
```diff
@@ -12,7 +12,9 @@
...
+  "STYLE is \"\", \"insensitive\" or \"smart\"")
```
Sarcasm: This line is not correct. I think it's `exact`, `case-insentitive` and `smart-case`.

server/src/Irony.cpp
```diff
@@ -286,6 +286,46 @@ void Irony::complete(const std::string &file,
   std::cout << "(success . t)\n";
 }
 
+namespace {
+
+bool hasUppercase(const std::string &prefix)
+{
+  for (auto it = prefix.begin(); it != prefix.end(); ++it) {
```
Sarcasm: Prefer range-based for loop:

```cpp
  for (char c : prefix) {
    if (std::isupper(c)) {
      return true;
    }
  }

  return false;
```

server/src/Irony.cpp
```diff
@@ -428,12 +471,28 @@ void Irony::candidates() const {
...
         // annotation is what comes after the typedtext
         annotationStart = prototype.size();
         typedTextSet = true;
       }
     }
 
+    if (!hasPrefix) {
+      continue;
+    }
+    if (!typedTextSet) {
```
Sarcasm: I'm not 100% sure to understand this situation, feel free to elaborate if you remember a minimal example.

server/src/Commands.def
kangjianbin: You are right. I forgot to update the document.

server/src/Irony.cpp
kangjianbin: I can reproduce it in clang3.9.1:
```c++
void foobar(int parameter)
{
}

int main(int argc, char *argv[])
{
    foobar(xxx <-complete here
}

```
Without the check, output of *Irony* (I disabled flycheck-mode) is:

```elisp
(
  ("" 1 "void" "" "foobar(int parameter)" 0 ("") available)
)

;;EOT

```

Sarcasm: Oh I see, this information could actually be useful to provide user hints, e.g. with eldoc.

Sarcasm:
It's merged!

Thank you very much for working on this.

netjune:
Where does the prefix come from? Does libclang give us too many unused candidates?

Issue #462, opened 2018-01-13
RPATH under macOS


hotpxl:
I'm using macOS High Sierra with MacPorts. libclang is installed in a non-default location so RPATH has to be set for the binary to work. I'm not sure how things are with people using Homebrew.

I wonder if we should enable by default, for macOS users, the CMake flag that sets RPATH?

PR #474, opened 2018-03-29
Preserve RPATH on macOS.

hotpxl:
This solves #462 
I'm using MacPorts, and the installer script of irony server would find the correct shared library, but removing rpath would cause it to become unusable.


Sarcasm:
Sorry, I don't think the build file is an appropriate place for this, it is more like a packaging/system issue.

However, I would be interested in finding the proper place for this.
Right now, the best place is for the user to edit the cmake command line when doing `M-x irony-install-server RET`, and to add `-DCMAKE_INSTALL_RPATH_USE_LINK_PATH=ON` to the command line, if it knows it make sense on his system.

hotpxl:
That makes sense as well. So are you suggesting having Emacs detect OS and put together compile flags?

Sarcasm:
That would also probably incorrect, it's not an OS-specific thing.
There is a plethora of methods to install Clang on macOS, and some of them needs this.

Sorry, for not being a very practical person.

Right now, I guess it could be a irony-mode configuration variable, e.g.: `irony-extra-cmake-args`.

If I wanted irony-mode to work "out of the box" for the major platforms, I would download a prebuilt LLVM release, put it in a directory only irony-mode uses, and set the appropriate flags if it needs it. But this comes with its own set of issues, so the user variable may be a better starting point. And it could be useful to specify other options anyway.

hotpxl:
I agree. I think ycm packages a libclang, cquery probably does it too. It's always painful.
I would agree to the extra config variable route. This way I can put it in my init.el and forget about it.

hotpxl:
I created PR #475 
How do you like that?

MaskRay:
@hotpxl Welcome to try https://github.com/MaskRay/ccls

MaskRay:
Deleted. hotpxl was beside me when I made that comment (as a joke). Deleted in case it makes others puzzled.

PR #475, opened 2018-04-06
Add extra arguments to CMake

hotpxl:
#474

### DIFF
--- a/irony.el
+++ b/irony.el
@@ -97,9 +97,9 @@ automatically buffer-local wherever it is set."
   :type 'string
   :group 'irony)
 
-(defcustom irony-extra-cmake-args ""
+(defcustom irony-extra-cmake-args nil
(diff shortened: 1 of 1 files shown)


Sarcasm:
I like it, with the name "irony-cmake-extra-args", I have in mind a list, like `irony-additional-clang-options`, e.g.:

```elisp
(setq irony-extra-cmake-args '("-G" "Ninja" "-DCMAKE_INSTALL_RPATH_USE_LINK_PATH=ON")
```

I don't really mind it's not a list, but I'm wondering if a more fitting name could be used then, any thoughts?

hotpxl:
Hmm. Ok I made a list (simpler than coming up with a new name :)

Sarcasm:
Yeah I had the same issue with finding another name...

Thanks for the patch!

PR #476, opened 2018-04-16
Make guess-flags choose the largest directory prefix

russellmcc:
It looks like this was the intended logic all along - changing it to
this makes the cdb-json more likely to guess the correct flags for
header files.


Sarcasm:
Indeed, good catch, thank you!

PR #486, opened 2018-06-24
Allow public interactive function to parse current buffer

3246251196:
There have been a number of people who have come across a situation where restarting the server would be useful. 

For instance, after a while the completion results can return trash and I have never looked into why this is yet!

Anyway, at work I restart irony-server every 120s. I needed to create a function to allow this to happen and I think other people would find it useful if creating emacs timers.

### DIFF
--- a/irony.el
+++ b/irony.el
@@ -902,11 +902,11 @@ old-state can be nil if the old state isn't known."
         (message "%s (aka '%s')" (car types) (cadr types))
       (message "%s" (car types)))))
 
-(defun irony-parse-buffer-async (callback)
-  "Parse the current buffer sending the results to CALLBACK."
-  (interactive)
+(defun irony-parse-buffer-async (&optional callback)
+  "Parse the current buffer sending results to an optional
+  CALLBACK function."
   (irony--run-task-asynchronously (irony--parse-task)
-                                  callback))
+                                  (or callback #'ignore)))
 
 (provide 'irony)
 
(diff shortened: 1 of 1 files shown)


3246251196:
I should mention that there is no need to add a public killing function since irony-server-kill already exists.

Sarcasm:
It's a bit sad we need such workarounds...but anyway.

I would rename the function `irony-parse-buffer-async`.
And it would make sense for the function to take the callback argument.
Then you are free to give a noop lambda in your config.

3246251196:
@Sarcasm 
I believe I have satisfied your request. The only thing I worry about is the user getting frustrated about needing to add a callback function. I think the user would find it easier to just call the function. I am wondering if there is anything like default argument values (as in C++) for lisp? So, we could default to an empty no-op lambda if no argument is provided. It would be trivial, I would guess: just make the `callback' an optional argument and check whether it has a value when inside the function, if it does not then provide the no-op lambda.

How do you feel about this? Or, is the latest version (a93a97b) good enough?

Thank you.

Sarcasm:
If you want callback to be an optional argument, you can do `&optional callback` in the parameter list.
Then, at the time of use, you can do:

```elisp
(or callback #'ignore)
```

Conveniently, `ignore` is a function that takes any number of arguments, and do nothing.

I don't think the function should be interactive, it will just clutter the list of interactive function provided when typing `Mx irony-|`.

3246251196:
@Sarcasm 
Thank you, I have applied the changes.

Sarcasm:
Merged, thank you.

Issue #384, opened 2017-05-03
clang_getCursorReferenced and clang_getCursorDefinition for jumping to declaration/definition


ikirill:
Since emacs has xref, it would be quite nice to expose `clang_getCursorReferenced` and `clang_getCursorDefinition`, so that `xref-find-definitions` would work. So you could jump to definition/declaration of whatever is under point. Are there any plans for this in irony-mode? Would you be interested in a PR?

(Plus, there's also `xref-find-references`, which is separate, but would also be nice.)

Sarcasm:
Yes, very much interested in a PR, and in using it. :)
irony-server won't be an a good indexer but for local stuff, jumping to the declaration of a function, it is a nice feature.

PR #504, opened 2018-10-29
Implement the main part of the xref interface

ikirill:
I left for a while and completely forgot #384.

For the xref interface see https://github.com/emacs-mirror/emacs/blob/master/lisp/progmodes/xref.el, and the lisp xref backend in https://github.com/emacs-mirror/emacs/blob/master/lisp/progmodes/elisp-mode.el

This implements `xref-find-{definitions,references}`. It's not quite perfect
- it will jump into system headers (not sure if this is good or not, I've been using it that way myself)
- it will resolve a call of an overloaded function to the right definition
- The `find-references` functionality seems a little wonky: it seems to find references to the thing at point only in the same file plus the file in which it is defined (?). Maybe I got the AST traversal logic wrong in `Irony::xrefReferences`, but I'm not sure.

There a several FIXME's left in the code, which may or may not be issues (please have a look), they were things I wasn't quite sure about.

I tested this in my own code, it seems to work, but I don't have any actual unit tests for it.

### DIFF
--- a/irony-xref.el
+++ b/irony-xref.el
@@ -143,5 +143,14 @@
 ;;      collect
 ;;      (xref-make name (xref-make-file-location filename line column)))))
 
+
+;; Setting up xref
+
+(defun irony-xref--enter ()
+  (add-hook 'xref-backend-functions #'irony--xref-backend nil t))
+
+(defun irony-xref--exit ()
+  (remove-hook 'xref-backend-functions #'irony--xref-backend t))
+
 (provide 'irony-xref)
 ;;; irony-xref.el ends here
--- a/irony-xref.el
+++ b/irony-xref.el
@@ -6,6 +6,11 @@
 ;;
 ;;     (add-hook 'irony-mode-hook (lambda () (add-hook 'xref-backend-functions #'irony--xref-backend nil t)))
 ;;
+;; Features:
+;; - will jump into system headers
+;; - with overloaded functions, will jump to the right function definition
+;;   (it's not just string matching on the function name)
+;;
 ;; Missing commands:
 ;; - ‘xref-find-apropos’
 ;;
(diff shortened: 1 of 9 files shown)


ikirill:
`xref-apropos` is missing because it is supposed to search for functions by string matching, and I couldn't find a way to make that into something meaningful specifically in C++.

irony.el
```diff
@@ -426,6 +428,7 @@ If no such file exists on the filesystem the special file '-' is
       (display-warning 'irony "Performance will be bad because a\
  pipe delay is set for this platform (see variable\
  `w32-pipe-read-delay')."))))
+  (irony-xref--enter)
```
Sarcasm: Should this be optional?
I think some people use irony for completion only, and sometime use rtags or something else for cross reference.

irony-xref.el
```diff
@@ -0,0 +1,140 @@
...
+(defun irony-find-references (&optional pos)
+  (interactive)
+  (let* ((line-column (irony--completion-line-column pos))
+         (result (irony--run-task
+                  (irony-iotask-chain
+                   (irony--parse-task) ; FIXME Is parsing necessary here?
+                   (irony-iotask-package-task irony--t-xref-references
+                                              (car line-column) (cdr line-column))))))
+    (cl-loop for item in result
+             do (message "%S" item))))
```
Sarcasm: debug?

server/CMakeLists.txt
```diff
@@ -14,6 +14,8 @@ include(CTest)
 check_for_in_source_build()
 release_as_default_build_type()
 
+set(CMAKE_CXX_FLAGS_DEBUG "${CMAKE_CXX_FLAGS_DEBUG} -g -O0 -fsanitize=address -fsanitize=undefined -Weverything -Wno-c++98-compat -Wno-c++98-compat-pedantic -Wno-padded -Wno-missing-prototypes -Wno-shadow-field-in-constructor")
```
Sarcasm: debug?

Issue #489, opened 2018-06-29
support compile_flags.txt


ensc:
Recent clang supports a `compile_flags.txt` file which can be used as an alternative to `compilation-database.json`. Quoting from https://clang.llvm.org/docs/JSONCompilationDatabase.html#build-system-integration

> ### Alternatives
> For simple projects, Clang tools also recognize a compile_flags.txt file. This should contain one flag per line. The same flags will be used to compile any file.

It would be nice when irony-mode would support this file.

PR #505, opened 2018-10-29
Add support for compile_flags.txt

ikirill:
See #489 

The format of compile_flags.txt appears to be compatible with the format of .clang_complete. If https://github.com/Rip-Rip/clang_complete has the right definition of the format, then it's the same one-flag-per-line type of thing.

### DIFF
--- a/irony-cdb-clang-complete.el
+++ b/irony-cdb-clang-complete.el
@@ -22,6 +22,11 @@
 ;;
 ;; This file defines a compilation database for .clang_complete file.
 ;;
+;; Another type of compilation database, compile_flags.txt
+;; (https://clang.llvm.org/docs/JSONCompilationDatabase.html#build-system-integration)
+;; appears to share a very similar format, so is processed the same
+;; way as a .clang_complete file.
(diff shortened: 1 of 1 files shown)


irony-cdb-clang-complete.el
```diff
@@ -40,8 +45,15 @@
 
 (defun irony-cdb-clang-complete--locate-db ()
   (when buffer-file-name
-    (irony--awhen (locate-dominating-file buffer-file-name ".clang_complete")
-      (concat (file-name-as-directory it) ".clang_complete"))))
+    (let ((clang-complete
```
Sarcasm: I think it would be better to have only one "locate-dominating-file".
We should traverse the filesystem only once, and the first of the 2 files should be returned.

Sarcasm:
Closed by #556

campisano:
@Sarcasm , should the Readme.md mention the compile_flags.txt together with the .clang_complete in the https://github.com/Sarcasm/irony-mode#compilation-database section?

PR #529, opened 2019-05-13
Search every subdir under /usr/lib{,64}/clang

tastytea:
`LIBCLANG_KNOWN_LLVM_VERSIONS` in [FindLibClang.cmake](https://github.com/Sarcasm/irony-mode/blob/master/server/cmake/modules/FindLibClang.cmake#L18) will never contain all possible versions. Using a wildcard is simpler.

### DIFF
--- a/server/cmake/CheckClangResourceDir.cmake
+++ b/server/cmake/CheckClangResourceDir.cmake
@@ -37,7 +37,8 @@ function(check_clang_resource_dir)
     "${LIBCLANG_LIBRARY_DIR}/"         # Google
     "/usr/lib64/clang"                 # x86_64 (openSUSE, Fedora)
     "/usr/lib/clang"
-    PATH_SUFFIXES ${LIBCLANG_KNOWN_LLVM_VERSIONS})
+    "/usr/lib64/clang/*"
+    "/usr/lib/clang/*")
 
   if (CHECK_CLANG_RESOURCE_DIR)
     # On Windows the paths weren't escaped correctly, similar to:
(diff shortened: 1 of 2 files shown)


Sarcasm:
Hello, thanks for the PR, however I'm not convinced by this fix.

I would go another route.
Recent versions of clang, should distribute an ClangConfig.cmake, which create the `libclang` IMPORTED target.
And the usual CMake arguments can be used to point to a specific version if needed.
However, when I tried to use this method, I found some issues, which I tried to fix: https://reviews.llvm.org/D30911
The initial discussion which shows what `find_package(Clang)` returns: https://github.com/Sarcasm/irony-mode/pull/368/files
I think the fix is in since Clang >= 5.1, but I worked on my patch since then.

So maybe you can take a look at using `find_package(Clang)`, or I should try that myself, still have the git stash around.

Also a few questions:
- where is it documented that `find_path( PATHS )` supports globbing? I could not find it here: https://cmake.org/cmake/help/v3.14/command/find_path.html
- does your method uses the most recent versions if multiples are available, or does it picks the first one in glob order, which appears to be the oldest one when globbing in my shell?

tastytea:
I could not find documentation that states that `find_path( PATHS )` supports globbing, but it works.

`list( APPEND )` has also no documented support for globbing. It does support globbing in cmake 3.9.6 on Gentoo but apparently not in cmake 3.9.2 on Ubuntu.

I would guess that my method picks the first (oldest) version that it finds.

In summary, I don't think my patch is a good idea anymore.

Sarcasm:
Ok, I will close the issue then
But thank you for bringing this up.

I made a patch to use `find_package(Clang)`:
- https://github.com/Sarcasm/irony-mode/commit/0ae10b7eaab6f76af5685ecb41e9400f25748085

The old code is still used as a fallback.
If the patch happens to work (I have some doubts), it will not be necessary to update the list of versions anymore.

PR #537, opened 2019-06-09
Look for irony-server in irony.el directory

matthewbauer:
This makes irony--find-server-executable look in
<load-file-name>/server/bin for the irony-server executable. This is
useful when you want the irony-server executable to be preinstalled
system-wide in your Emacs setup. This saves users the trouble of
having to individually build irony-server and have all of the
dependencies that come with it.

### DIFF
--- a/irony.el
+++ b/irony.el
@@ -570,7 +570,10 @@ The installation requires CMake and the libclang developpement package."
 Throw an `irony-server-error' if a proper executable cannot be
 found."
   (let* ((exec-path (cons (expand-file-name "bin" irony-server-install-prefix)
-                          exec-path))
+                          (cons (expand-file-name "server/bin"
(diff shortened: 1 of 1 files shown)


Sarcasm:
Do you have a package manager in mind that does this, or some other Emacs module?

For a system package manager, I would expect irony-server to end up in the user PATH eventually, which I think already works.
I'm not sure how and where Elpa packages are supposed to install binaries or Emacs modules, but I don't really want to invent multiple conventions of my own.

matthewbauer:
Yeah, just putting it in PATH makes sense. ELPA doesn't really have a way to handle executables though, so it's a little bit weird. In Nixpkgs, we just put everything into ```/share/emacs/site-lisp/elpa/<pkg>```. But I think it can be changed to put it in ```/bin``` instead.

Issue #543, opened 2019-07-06
irony server installation


aviikc:
0

Irony install server yeilds this error every time i try an installation. Clang requirements are all installed. Can i be helped about what I'm missing here.

Can someone please help me out. 
-*- mode: compilation; default-directory: "/tmp/build-irony-server-1.3.1/" -*-
Compilation started at Sat Jul  6 08:23:45
 
cmake -DCMAKE_INSTALL_PREFIX\=/home/aviik/.emacs.d/irony/  /home/aviik/.emacs.d/elpa/irony-20190703.1732/server && cmake --build . --use-stderr --config Release --target install
-- The C compiler identification is GNU 6.3.0
-- The CXX compiler identification is GNU 6.3.0
-- Check for working C compiler: /usr/bin/cc
-- Check for working C compiler: /usr/bin/cc -- works
-- Detecting C compiler ABI info
-- Detecting C compiler ABI info - done
-- Detecting C compile features
-- Detecting C compile features - done
-- Check for working CXX compiler: /usr/bin/c++
-- Check for working CXX compiler: /usr/bin/c++ -- works
-- Detecting CXX compiler ABI info
-- Detecting CXX compiler ABI info - done
-- Detecting CXX compile features
-- Detecting CXX compile features - done
CMake Error at /usr/lib/llvm-9/lib/cmake/clang/ClangTargets.cmake:581
 (message):
  The imported target "clang_shared" references the file 
     "/usr/lib/llvm-9/lib/libclang_shared.so.9"
 but this file does not exist.  Possible reasons include:

* The file was deleted, renamed, or moved to another location.

* An install or uninstall procedure did not complete successfully.

* The installation package was faulty and contained

"/usr/lib/llvm-9/lib/cmake/clang/ClangTargets.cmake"

but not all the files it references. 
Call Stack (most recent call first):
  /usr/lib/cmake/clang-9/ClangConfig.cmake:19 (include)
  src/CMakeLists.txt:4 (find_package)

-- Configuring incomplete, errors occurred!
See also "/tmp/build-irony-server-1.3.1/CMakeFiles/CMakeOutput.log".

[CMakeOutput.log](https://pastebin.com/JT1TJFjc)

Sarcasm:
What is your OS/distribution?
Looks like a packaging issue to me.

aviikc:
i'm using Debian GNU/Linux 9 (stretch)

Sarcasm:
[...]

PR #552, opened 2019-12-04
README: Add section about how to access newer versions of Clang/LLVM

sten0:
Hi @Sarcasm,

The motivation for this PR is to document the best methods for Debian (and derivatives) users to use irony-mode with a newer version of Clang, both the officially supported and unsupported method.  I discussed #543 with the Sylvestre Ledru, and he recommends these users use his LLVM nightly repository, in case the stable version provided in backports wasn't new enough.  If you'd prefer not to merge it I'd be happy to maintain a patch against the Debian (and derivatives) package.  Please let me know if you prefer differently styled formatting, or feel free to make modification.

Cheers,
Nicholas

### DIFF
--- a/README.md
+++ b/README.md
@@ -56,6 +56,20 @@ package manager.
 * Using `apt` on Debian ≥10 and derivatives
 
         sudo apt install elpa-irony
+#### Accessing a newer Clang and LLVM Toolchain on Debian and derivatives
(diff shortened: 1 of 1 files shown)


Sarcasm:
LGTM, thanks!

PR #556, opened 2020-01-24
Support for compile_flags.txt

### LINKED ISSUE #489: support compile_flags.txt

vaivaswatha:
This addresses the comment in #505, which looks to be abandoned.

Closes #489.

### DIFF
--- a/irony-cdb-clang-complete.el
+++ b/irony-cdb-clang-complete.el
@@ -20,7 +20,8 @@
 
 ;;; Commentary:
 ;;
-;; This file defines a compilation database for .clang_complete file.
+;; This file defines a compilation database for .clang_complete and
+;; compile_flags.txt, both of which have the same format.
 ;;
 
 ;;; Code:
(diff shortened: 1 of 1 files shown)


Sarcasm:
Thanks!
It's slightly better than the previous one, but there is still 2 traversals if the user is using `compile_flags.txt`.
I'm pretty one call to `locate-dominating-file` is sufficient, can you try that?

vaivaswatha:
> Thanks!
> It's slightly better than the previous one, but there is still 2 traversals if the user is using `compile_flags.txt`.
> I'm pretty one call to `locate-dominating-file` is sufficient, can you try that?

Done.

Sarcasm:
Just a thought, as I feel like this could simplify the code even further (but could be wrong).

Can you `throw` from the `locate-dominating-file` function, and return the value if you catch any?
Also, I would favor looking for `.compile_flags.txt` first, as it is now a somewhat standard file, documented by Clang itself.

If you don't think it's worth it, I'm ok merging the code as-is.

vaivaswatha:
> Just a thought, as I feel like this could simplify the code even further (but could be wrong).
> 
> Can you `throw` from the `locate-dominating-file` function, and return the value if you catch any?
> Also, I would favor looking for `.compile_flags.txt` first, as it is now a somewhat standard file, documented by Clang itself.
> 
> If you don't think it's worth it, I'm ok merging the code as-is.

Pushed both the changes you suggested.

Sarcasm:
Great!
I think the last thing that could have been improved is the double-ifs.
Using a `when`, `cond` or maybe `pcase` could make this part slightly more succinct.

Thank you for your time and effort. :+1:

PR #557, opened 2020-02-13
Add 10.0.0 to LIBCLANG_KNOWN_LLVM_VERSIONS

rosslannen:


Sarcasm:
Since https://github.com/Sarcasm/irony-mode/releases/tag/v1.4.0 this sort of changes should no longer be necessary.

If irony-mode was unable to pick up the LLVM CMake installation, it would be nice to know why.
What error do you have when you don't add this version?
Can you add some debug around here:

https://github.com/Sarcasm/irony-mode/blob/5f75fc0c9274f4622470e2324e2f4457087aa643/server/src/CMakeLists.txt#L4-L23

Also, the issue could be in your LLVM installation.

rosslannen:
I apologize for taking so long to get back. An upgrade to a newer version of irony fixed my issue. Thanks!

PR #565, opened 2020-07-31
Changed the dead melpa link with the new one.

RushanKhan1:
The MELPA hyperlink (http://melpa.milkbox.net/) is dead so I replaced it with the current one (https://melpa.org/#/).


Sarcasm:
Thank you!

PR #569, opened 2020-10-15
Revert change to Irony mode to patch a now fixed bug in Clang that has not existed since about version 9

3246251196:
I finally have got around to removing the fix introduced to irony to display non-accessibles. Recall that before version 9 of LLVM/Clang (or around that time) **protected** members/methods were not showing in derived classes in completions due to a bug in LLVM/Clang.

### DIFF
--- a/irony-completion.el
+++ b/irony-completion.el
@@ -259,8 +259,10 @@ A candidate is composed of the following elements:
     more indices. These indices work by pairs and describe ranges
     of placeholder text.
     Example: (\"(int a, int b)\" 1 6 8 13)"
-  (and (irony-completion-candidates-available-p)
-       irony-completion--candidates))
+  (irony--awhen (irony-completion-symbol-bounds)
+    (irony-completion--filter-candidates
+     (irony--run-task
+      (irony--candidates-task nil (car it) prefix style)))))
 
 (defun irony-completion-candidates-async (callback &optional prefix style)
   (irony--aif (irony-completion-symbol-bounds)
--- a/irony-completion.el
+++ b/irony-completion.el
@@ -57,19 +57,6 @@
   :type '(repeat function)
   :group 'irony-completion)
 
-(defcustom irony-completion-availability-filter '(available deprecated)
- "For completion, only accept candidates whose availability is in the list.
-
-Maps to libclang's CXAvailabilityKind:
-- https://clang.llvm.org/doxygen/group__CINDEX.html#gada331ea0195e952c8f181ecf15e83d71
-
-Due to a bug in
-Clang (https://bugs.llvm.org//show_bug.cgi?id=24329), candidates
-that can be validly accessed are deemed not-accessible."
- :type '(repeat symbol)
- :options '(available deprecated not-accessible)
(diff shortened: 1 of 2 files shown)


3246251196:
See #369

irony-completion.el
```diff
@@ -234,10 +221,6 @@ displayed when a derived class overrides virtual methods."
 (defun irony-completion-post-comp-placeholders (candidate)
   (cdr (nth 6 candidate)))
 
-(defun irony-completion-availability (candidate)
```
Sarcasm, line 237: This has been removed, but it's still used by `irony-completion--filter-candidates` below.

irony-completion.el
```diff
@@ -276,12 +259,9 @@ A candidate is composed of the following elements:
...
     more indices. These indices work by pairs and describe ranges
     of placeholder text.
-    Example: (\"(int a, int b)\" 1 6 8 13)
- 7. The availability (CXAvailabilityKind) of the candidate." 
-  (irony--awhen (irony-completion-symbol-bounds)
-    (irony-completion--filter-candidates
-     (irony--run-task
-      (irony--candidates-task nil (car it) prefix style)))))
+    Example: (\"(int a, int b)\" 1 6 8 13)"
+  (and (irony-completion-candidates-available-p)
```
Sarcasm, line 263: Is this code correct?
Looks like it's not only the documentation that changed, but the implementation too, the function is no longer running a task so I'm a bit surprised.

Sarcasm (changes requested):
Thank you for following up on this bug.

Sarcasm:
Although the commit is named "bad1285", I think it's a nice addition. :grinning: 
Thank you!

3246251196:
@Sarcasm,

I definitely think we should revert the commit. I was not happy with it.

The fact is I will never get my head around git. I have been an svn user for years and I just cannot get used to this thing. So I was making local changes on my local machine due to your review comments. Then I pushed the changes not realising they would show up here and that you would get updated. 

I think what you have done is trusted my changes too much. I have not even tested thi which I plan on doing. And I was relying on the automatic tests that normally used to run when I made a pull request. 

Are you actually happy and have you tested irony-nodebas it is with this latest commit because I fear that git is trying to tell us something with "bad" in the ID.

Thanks.

Sarcasm:
Oh, I thought you tested it locally before creating a pull request.
The initial change didn't seem working to me (removing the run-task), but you fixed that and the commit looked reasonably safe to me.
There is not much automated tests, and they usually don't test Clang, but more like helper functions.

I'm okay to revert, but what are you afraid off to be broken?
You mostly removed special cases here and there.

Before I revert, you can maybe test it a bit, and see if completion isn't broken, and that inaccessible (private?) fields aren't offered.

3246251196:
Hi @Sarcasm,

Let me do some testing and I'll update you later. At the latest on Sunday night. If you hear nothing then you should revert and I will perform another pull request from fresh. If it works then I will let you know. Sorry, I should have tested locally before and never realised that git push would update you. 

Let's talk soon.

PR #584, opened 2021-12-10
Clarify Debian installation instructions

sten0:
Also, clarify the recommendation to not use multiple package managers
to install multiple copies of irony-mode.

Addresses issues 516 and 583


sten0:
Force pushed for fixups (that you would have asked to squash), and then because I accidentally rebased onto the Debian remote. :$

Sarcasm:
Thanks for the clarifications!

PR #586, opened 2022-07-19
Add option to disable when editing over tramp

richyliu:
This PR adds the ability to disable irony-mode when editing files over tramp (to prevent errors when irony tries to access files over tramp).

```elisp
(setq irony-disable-over-tramp t)
```


Sarcasm:
Not sure this needs to be built into irony mode, user configuration can filter-out tramp just as well, no?

richyliu:
You're right, I just realized I can disable it in a hook myself.

